PR #15, opened 2025-11-20
chore: manage node/bun version with pnpm

ryoppippi:
in the latest pnpm, we can manage node/bun in pnpm. which is useful


steipete (commented):
Thank you, that's a nice cleanup. Merged.

PR #17, opened 2025-11-22
ci: use matrix and update actions

ryoppippi:
use matrix to remove duplicated actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,8 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for cleaning up!

steipete (commented):
Summary:\n- Nice consolidation using a single matrix job for linux/macos/windows.\n\nConcerns/questions:\n- The workflow now uses actions/checkout@v6, which is still labeled beta and requires runner >= v2.329.0; consider staying on the stable v5 until v6 is GA to avoid surprise runner incompatibility.\n- The test step no longer sets FIRECRAWL_API_KEY and LINEAR_API_KEY to dummy values; if any config interpolation expects them, tests could start reading real secrets or fail when missing. If they’re unused, all good—otherwise keep the explicit test defaults.

PR #27, opened 2025-12-17
fix: coerce array elements to proper types based on JSON schema

rawwerks:
## Summary

When parsing CLI array arguments (e.g., `--coords 1.5,2.5`), the values were split by comma but kept as strings. This caused validation failures for MCP tools that expect typed arrays like `number[]` or `boolean[]`.

## Changes

- Add `arrayItemType` field to `GeneratedOption` to capture the array's item type from the JSON schema
- Add `inferArrayItemType()` helper to extract item type from schema's `items.type`
- Update `optionParser()` to coerce array elements based on item type:
  - `number`: `parseFloat()` each element
  - `boolean`: compare against `'false'`
  - `string`/default: trim whitespace

## Example

For a tool with schema:
```json
{
  "coordinates": {
    "type": "array",
    "items": { "type": "number" }
  }
}
```

Before: `--coordinates 1.5,2.5` → `["1.5", "2.5"]` (strings, fails validation)
After: `--coordinates 1.5,2.5` → `[1.5, 2.5]` (numbers, passes validation)

## Testing

Tested with MCP servers that have array parameters expecting numbers - the generated CLI now correctly passes typed arrays to the underlying tools.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ## [Unreleased]
 
-No unreleased changes yet.
+### CLI
+- Coerce generated CLI array arguments based on JSON Schema item types (including integer arrays).
+
+### Tests
+- Added regression coverage for typed array parsing in generated CLIs.
 
 ## [0.7.2] - 2025-12-29
 ### CLI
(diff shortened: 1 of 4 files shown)


steipete:
Thanks! Taking over.

PR #49, opened 2026-01-31
fix: collect JSON from multiple content blocks into an array

Agasper:
## Summary
- When an MCP server returns `list[BaseModel]` (e.g. FastMCP serializing a Python list of Pydantic models), each item is serialized as a separate text content block in the MCP response
- Previously, `json()` in `createCallResult` returned on the first successfully parsed JSON entry, silently discarding the rest
- This caused tools like `yandex-tracker-mcp.issues_find` (which returns `list[Issue]`) to appear as if they only returned 1 result when there were actually hundreds

**The fix:**
- Collects all valid JSON entries from content blocks
- Single entry → returns the object directly (no behavior change)
- Multiple entries → returns them as an array

## Test plan
- [x] All 18 result-utils tests pass
- [x] Added 3 new test cases covering multi-content JSON collection:
  - `collects json from multiple text content blocks into an array`
  - `returns single object (not array) when only one content block has json`
  - `collects json from mixed json and text content blocks`

### DIFF
--- a/src/result-utils.ts
+++ b/src/result-utils.ts
@@ -169,13 +169,17 @@ export function createCallResult<T = unknown>(raw: T): CallResult<T> {
 
       const content = extractContentArray(raw);
       if (content) {
+        // When multiple content blocks each contain valid JSON (e.g. FastMCP
+        // serialising list[BaseModel] as separate text entries), collect all
+        // parsed values and return them as an array instead of only the first.
+        const collected: unknown[] = [];
         for (const entry of content) {
           if (entry && typeof entry === 'object') {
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for this fix. We’re closing as superseded because this behavior is already landed in `main` via PR #91.

Replacement:
- PR #91: https://github.com/steipete/mcporter/pull/91
- Merge commit: cb0639273864d837ef9508bb8a50ef15aa20f434

Current `main` already collects all parseable JSON entries from MCP content arrays and has regression coverage in `tests/result-utils.test.ts`.

Really appreciate the contribution.

PR #91, opened 2026-03-02
fix: collect all JSON entries from MCP content arrays instead of only the first

Blankdlh:
## Problem

When an MCP server returns a `content` array with multiple items (e.g., a list of resources or results), `json()` in `result-utils.ts` only returns the **first** parseable JSON entry — all subsequent items are silently discarded.

This happens because the `for...of` loop in `json()` uses an early `return` on the first successful parse:

```ts
for (const entry of content) {
  // ...
  const parsed = tryParseJson(entry);
  if (parsed !== null) {
    return parsed as J;  // ← exits immediately, remaining entries are lost
  }
}
```

## Fix

- **`result-utils.ts`**: Changed `json()` to collect all parseable entries into an array. Returns a single object when only one item exists (backward compatible), or an array when multiple items are found.
- **`output-utils.ts`**: Increased `inspect` depth in `printRaw()` from `2` to `null` (unlimited) to prevent truncation of deeply nested list results.

## Tests

Added 2 new test cases to `result-utils.test.ts`:

- Multiple JSON-type content entries → returns all as an array
- Mixed JSON + text-parseable-as-JSON entries → returns all as an array

All 17 tests pass (15 existing + 2 new).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Preserve default imports when `mcporter config add` writes a config file, instead of forcing `"imports": []`.
 - OAuth: avoid crashing on headless Linux when `xdg-open` is unavailable; clear stale dynamic-port client registrations; close callback server if stale-client persistence reads fail. (PR #72, thanks @mgonto)
 - Added optional `oauthScope`/`oauth_scope` config override as an escape hatch for providers that require explicit scopes.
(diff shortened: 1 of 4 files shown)


steipete:
Landed via temp rebase onto main.

- Gate: pnpm check && pnpm test && pnpm build && pnpm run docs:list
- Land commit: fd145326d60049bfee56523d7234da4b6295c0d0
- Merge commit: cb0639273864d837ef9508bb8a50ef15aa20f434

Thanks @Blankdlh!

Issue #90, opened 2026-03-01
config add silently drops --args flags from written config


miguelarios:
`mcporter config add` accepts `--args` flags without error and reports success, but the written mcporter.json file contains only the command field — all args are silently dropped.

I am executing this:

```
mcporter config add youtube-transcript \
  --command "uvx" \
  --args "--from" \
  --args "git+https://github.com/jkawamoto/mcp-youtube-transcript" \
  --args "mcp-youtube-transcript" \
  --args "--response-limit" \
  --args "15000"
```

and it is writing this:

```
{
  "mcpServers": {
    "youtube-transcript": {
      "command": "uvx"
    }
  },
  "imports": []
}
```

I'd expect this

```
{
  "mcpServers": {
    "youtube-transcript": {
      "command": "uvx",
      "args": [
        "--from",
        "git+https://github.com/jkawamoto/mcp-youtube-transcript",
        "mcp-youtube-transcript",
        "--response-limit",
        "15000"
      ]
    }
  },
  "imports": []
}
```

mcporter response `Added 'youtube-transcript' to /app/config/mcporter.json`

steipete:
Fixed on main by PR #93. mcporter config add now accepts repeated --args as an alias for stdio args, and tests/config-add-flags.test.ts covers preserving both values in the written entry. Closing as completed.

PR #93, opened 2026-03-03
fix: support --args flag (plural) in config add

Jah-yee:
Issue #90: config add silently drops --args flags

The CLI only handled --arg (singular) but users may pass --args (plural).
Now both --arg and --args are accepted.


steipete:
Landed the useful part directly on main in 97ac209863196a40e3201b639166dd7e18514f51.

Kept:
- --args alias in mcporter config add
- regression coverage for the alias and updated config-CLI error expectation

Not merged from this branch:
- unrelated binary artifact
- unrelated output-utils changes

Thanks @Jah-yee!

PR #104, opened 2026-03-12
fix: reconnect with fresh transport after OAuth completion

KentonYu:
After a successful OAuth browser flow, `connectWithAuth` looped back to `client.connect(transport)` reusing the same StreamableHTTPClientTransport instance.  The MCP SDK does not allow calling `start()` twice on a transport, so it threw "StreamableHTTPClientTransport already started!".

Because this error was neither an `UnauthorizedError` nor an `OAuthTimeoutError`, the catch block in `createClientContext` fell through to the SSE fallback.  Servers that only support Streamable HTTP (POST) then returned 404 on the SSE GET request, producing a confusing "SSE error: Non-200 status code (404)" — even though OAuth had already succeeded and tokens were persisted.

The fix introduces `OAuthCompletedError`: after `finishAuth` succeeds, `connectWithAuth` throws this sentinel instead of retrying on the stale transport.  `createClientContext` catches it and `continue`s the outer loop, which creates a brand-new Client + Transport with the now-cached tokens and connects cleanly.

### DIFF
--- a/src/runtime/transport.ts
+++ b/src/runtime/transport.ts
@@ -149,6 +149,8 @@ export async function createClientContext(
       } catch (primaryError) {
         if (primaryError instanceof OAuthCompletedError) {
           // OAuth succeeded but the transport is already started; retry with a fresh transport.
+          // Close the current session's callback server to free the port before the next iteration.
+          await oauthSession?.close().catch(() => {});
           logger.info(`OAuth complete for '${activeDefinition.name}'. Reconnecting...`);
           continue;
         }


KentonYu:
@steipete  Please help me review this PR. When we use mcporter auth, we can't use MCP services like notion normally

nabigraphics:
@KentonYu one follow-up from my testing: `catch (sseError)` may also need the same `OAuthCompletedError` handling.

I reproduced this with Atlassian MCP when the initial HTTP path returned `502` and `mcporter` fell back to SSE. Since the SSE fallback path also goes through `connectWithAuth()`, it can raise the same `OAuthCompletedError` after OAuth completes successfully. Handling that sentinel in the SSE catch path made the Atlassian auth flow complete successfully on my side.

steipete:
Closing as superseded by #97. Main now reconnects with a fresh transport after OAuth completion and preserves the post-auth failure context, so this older sentinel-based approach is no longer needed. Thanks for the clear root-cause write-up.

PR #107, opened 2026-03-13
Saved

lehuubiensea25122003-top:


steipete:
seems garbage, no text

steipete:
Triage note: already closed. Keeping closed rather than landing this generated performance branch as-is: it includes generated summary docs and a binary dist artifact, and the implementation no longer matches current main. We can extract a focused benchmark/cache change later. Thanks!

PR #111, opened 2026-03-16
docs: fix incorrect top-level heading rendering

polaris-dxz:


steipete:
Closing as not needed: this only adds whitespace in AGENTS.md and does not materially improve rendering in current docs. Thanks!

PR #124, opened 2026-03-20
fix: handle resource content blocks in output rendering

### LINKED ISSUE #108: resource content type blob/text silently dropped in all --output modes
## Bug

When an MCP tool returns a `{ type: "resource", resource: { uri, mimeType, text } }` content item, the resource text/blob is silently dropped across all `--output` modes. The result is either empty output or a truncated `[Object]` placeholder with no accessible content.

## Reproduction

Using the GitHub MCP server (`github.get_file_contents`):

```bash
# plain output
mcporter call github.get_file_contents owner=myorg repo=myrepo path=docs/README.md
# Output: "successfully downloaded text file (SHA: abc123)"
# File content: MISSING

# --output json
mcporter call github.get_file_contents owner=myorg repo=myrepo path=docs/README.md --output json
# Output: { content: [ { type: 'text', text: 'successfully downloaded...' }, { type: 'resource', resource: [Object] } ] }
# resource blob: [Object] -- Node.js truncation, content unreachable
```

The actual MCP server response (verified by calling the SSE endpoint directly with curl) contains:
```json
{
  "type": "resource",
  "resource": {
    "uri": "repo://org/repo/sha/.../path",
    "mimeType": "text/plain; charset=utf-8",
    "text": "...full file contents here..."
  }
}
```

## Workaround

Calling the SSE endpoint directly with curl bypasses mcporter and returns full content.

## Expected Behavior

- `--output text` / `--output markdown`: render `resource.text` or decode `resource.blob`
- `--output json` / `--output raw`: include the full resource object, not a Node.js `[Object]` truncation

## Environment
- mcporter 0.7.3, Node v22.22.1, Linux x64
- GitHub MCP server via HTTP/SSE transport

## Root Cause Hypothesis

Output renderers likely only handle `type: "text"` and `type: "image"`. The `type: "resource"` content items appear to have no renderer -- the `text`/`blob` fields inside are never extracted. The `--output raw` path compounds this by using Node.js object serialization which truncates nested objects as `[Object]`.

mvanhorn:
## Summary

MCP tools returning `type: 'resource'` content blocks now render correctly instead of being silently dropped.

## Why this matters

Tools like `github.get_file_contents` return resource content blocks with `{ type: 'resource', resource: { uri, mimeType, text } }`. The `collectCallContent()` function in `result-utils.ts` only handled `json`, `text`, `image`, and `markdown` types - resource blocks hit the catch-all `continue` at line ~119 and were silently skipped.

## Changes

- `src/result-utils.ts`: Added resource content block handler before the catch-all guard. Extracts `resource.text` for text-based resources and creates a URI placeholder for binary resources. JSON-parseable resource text is also added to `jsonCandidates`.
- `tests/result-utils.test.ts`: Added test cases for text resources, binary resources, mixed content blocks, and JSON extraction from resource text.

Fixes #108

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### CLI
+- Render `resource` content blocks in call output helpers instead of dropping them, including markdown resources and JSON text payloads. (PR #124, thanks @mvanhorn)
 - Preserve default imports when `mcporter config add` writes a config file, instead of forcing `"imports": []`.
 - OAuth: avoid crashing on headless Linux when `xdg-open` is unavailable; clear stale dynamic-port client registrations; close callback server if stale-client persistence reads fail. (PR #72, thanks @mgonto)
 - Added optional `oauthScope`/`oauth_scope` config override as an escape hatch for providers that require explicit scopes.
--- a/src/result-utils.ts
+++ b/src/result-utils.ts
@@ -116,6 +116,22 @@ function collectCallContent(raw: unknown): CollectedCallContent {
       }
       continue;
     }
+    if (typedEntry.type === 'resource') {
+      const resource = typedEntry.resource as Record<string, unknown> | undefined;
+      if (resource && typeof resource === 'object') {
+        const uri = typeof resource.uri === 'string' ? resource.uri : '';
+        if (typeof resource.text === 'string') {
+          textEntries.push(resource.text);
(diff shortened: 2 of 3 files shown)


mvanhorn:
Thanks for the merge, @steipete!

mvanhorn:
Thanks for the merge!

PR #125, opened 2026-03-21
fix: add managed runtime and dedupe keep-alive restarts

zm2231:
## Summary
- add a public `createManagedRuntime()` helper that mirrors the CLI's keep-alive daemon wiring for library consumers
- switch the CLI to use `createManagedRuntime()` instead of duplicating daemon setup internally
- dedupe concurrent keep-alive restarts per server in `KeepAliveRuntime`
- document the managed runtime path in the README and add targeted tests

## Why
There are two related issues here:

1. The CLI already applies keep-alive daemon routing, but library users calling `createRuntime()` directly do not get equivalent behavior even though the README points them at that API for long-lived runtime usage.
2. `KeepAliveRuntime.invokeWithRestart()` retries fatal transport failures by calling `closeServer()` and retrying once, but concurrent failures against the same keep-alive server can race and tear down a freshly reconnected transport.

This PR keeps `createRuntime()` unchanged for callers that explicitly want direct runtime connections, while exposing a first-class managed runtime path that matches CLI semantics. It also makes restart handling idempotent per server so concurrent failures share the same in-flight restart.

## Validation
- `pnpm check`
- `pnpm exec vitest run tests/keep-alive-runtime.test.ts tests/runtime-managed.test.ts`

### DIFF
--- a/tests/cli-managed-runtime.test.ts
+++ b/tests/cli-managed-runtime.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it, vi } from 'vitest';
+
+process.env.MCPORTER_DISABLE_AUTORUN = '1';
+process.env.MCPORTER_NO_FORCE_EXIT = '1';
+
+const runtime = {
+  getDefinitions: vi.fn(() => []),
+  close: vi.fn(async () => undefined),
+};
+
+const createManagedRuntimeMock = vi.fn(async () => runtime);
+const createRuntimeMock = vi.fn(async () => runtime);
+const handleListMock = vi.fn(async () => undefined);
+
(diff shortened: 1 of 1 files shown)


steipete:
Landed via temp rebase onto main.

- Gate: ./runner pnpm docs:list && ./runner pnpm check && ./runner pnpm build && ./runner pnpm build:bun && ./runner pnpm test
- Land commit: b554ab399bd49859bd9a0aaa27ab1bb802318b5c
- Merge commit: 8c66f1c49a682226587f7fe8fee0be4264dcbfae

Squashed on merge to keep the branch cleanup and API-scope trim out of main history.

Thanks @zm2231!

PR #143, opened 2026-04-21
fix(oauth): log manual-completion URL at warn level so headless users can copy it

### LINKED ISSUE #139: Headless auth: browser URL for manual completion is missing in terminal and never shown, blocking server/CI usage
## Summary

When running `mcporter auth <server>` on a headless server, there is no way to see the required OAuth URL in order to complete the flow. The following steps were run on mcporter version 0.8.1 (latest as of this report):

```
mcporter config add datadog-mcp --url https://mcp.datadoghq.com/api/unstable/mcp-server/mcp --transport http
[mcporter] Added 'datadog-mcp' to /home/node/config/mcporter.json
mcporter --version
0.8.1
mcporter auth datadog-mcp
[mcporter] OAuth authorization required for 'datadog-mcp'. Waiting for browser approval...
```

At this point, **the CLI sits and waits for browser approval, BUT NO BROWSER URL IS PRINTED**. The log message that would include the OAuth URL does not appear at any point, even if the process is started in a typical terminal environment. The only output is the 'waiting for browser approval' message (which is logged at `warn` level).

## What should happen
- The CLI should print the OAuth authorization URL to stdout/stderr (or both), making it copyable, whenever code is waiting for browser approval. Ideally, this should happen regardless of logger configuration, shell, or CI environment.
- A flag such as `--headless` or `--print-url` should be added to guarantee this behavior explicitly if possible.

## Why this matters
- Without the URL, it is impossible to complete the flow in a headless or server context, because the user has nothing to copy-paste into a local browser.
- There is no way to work around this except (possibly) by patching the code or modifying logger implementation.
- This problem exists even though the code appears to attempt to log the URL (at `info` level); for some reason, users (including myself) do not see this line, only the warning.

## Additional context
- Logger code is using the same logger instance for both the warning and info lines. If info output is suppressed/missing, there is no fallback for the user.
- This occurs regardless of using `npx` or a global install, and regardless of shell redirection or TTY mode.
- See previously reported related needs in [#73](https://github.com/steipete/mcporter/issues/73), but this is a specific usability blocker for all headless environments where info logs are not appearing.

**Expected behavior:** The actual OAuth URL is always printed, reliably, when auth is waiting, especially when running in server or CI/headless mode, and ideally under an explicit flag.

stainlu:
Closes #139.

## Root cause

`logging.ts` defaults the log threshold to `warn`. The fallback URL line
in `redirectToAuthorization` is logged at `info`, so it silently drops
for every headless / CI user running with the default config:

| file:line | message | level |
|---|---|---|
| `src/oauth.ts:256` | `Authorization required for X. Opening browser...` | `info` (suppressed) |
| `src/oauth.ts:259` | `If the browser did not open, visit <url> manually.` | `info` (suppressed) |
| `src/runtime/oauth.ts:107` | `OAuth authorization required... Waiting for browser approval...` | `warn` (shown) |

The user sees only the "waiting" message and is stranded -- exactly the
symptom reported in #139 and corroborated by [@gaieges](https://github.com/gaieges).

## Fix

Elevate only the fallback-URL line to `warn`. That line is actionable
instruction (without it, a headless user cannot complete the flow), not
diagnostic status. The adjacent `"Opening browser..."` line stays at
`info` because it is purely informational.

Regression test in `tests/oauth-session.test.ts` asserts `logger.warn`
receives the URL-containing line.

## Intentionally out of scope (follow-ups)

- **`--headless` / `--print-url` flag.** The reporter asked for an
  explicit opt-in. Warn elevation covers the default-config case, which
  is what strands new headless users today. A fully level-independent
  path (direct `process.stderr.write`) would also help anyone running
  `MCPORTER_LOG_LEVEL=error`; happy to send that as a follow-up if you'd
  like.
- **`--browser none` documentation vs implementation.** `docs/config.md:139`
  promises `--browser none` suppresses the browser launch, but `grep`
  across `src/` finds no implementation -- flagged by @gaieges on the
  issue thread. Worth filing as its own bug (docs-only or implement the
  flag).

## Verification

- `pnpm check` -- format, oxlint, tsgo clean on touched files
- `pnpm test` -- 469 passed, 3 skipped (pre-existing), 0 failed
- Targeted: `pnpm exec vitest run tests/oauth-session.test.ts` -- 7 passed

### DIFF
--- a/src/oauth.ts
+++ b/src/oauth.ts
@@ -256,7 +256,10 @@ class PersistentOAuthClientProvider implements OAuthClientProvider {
     this.logger.info(`Authorization required for ${this.definition.name}. Opening browser...`);
     this.ensureAuthorizationDeferred();
     __oauthInternals.openExternal(authorizationUrl.toString());
-    this.logger.info(`If the browser did not open, visit ${authorizationUrl.toString()} manually.`);
+    // warn level because this is the manual-completion fallback: on headless servers / CI
+    // where the browser never opens, the user has no other way to recover the URL, so it
+    // must survive the default 'warn' log threshold set by logging.ts.
+    this.logger.warn(`If the browser did not open, visit ${authorizationUrl.toString()} manually.`);
   }
 
   async saveCodeVerifier(codeVerifier: string): Promise<void> {
--- a/tests/oauth-session.test.ts
+++ b/tests/oauth-session.test.ts
@@ -213,4 +213,43 @@ describe('FileOAuthClientProvider session lifecycle', () => {
     await expect(waitPromise).resolves.toBe('stable-deferred-code');
     await session.close();
   });
+
+  it('logs the manual-completion URL at warn level so headless/CI users can copy it (#139)', async () => {
+    const tokenCacheDir = await fs.mkdtemp(path.join(os.tmpdir(), 'mcporter-oauth-test-'));
+    tempDirs.push(tokenCacheDir);
+    const definition: ServerDefinition = {
+      name: 'test-oauth-headless-url',
+      description: 'Test OAuth server',
+      command: { kind: 'http', url: new URL('https://example.com/mcp') },
+      auth: 'oauth',
+      tokenCacheDir,
+    };
+    const logger = {
+      info: vi.fn(),
+      warn: vi.fn(),
+      error: vi.fn(),
+    };
+
+    const session = await createOAuthSession(definition, logger);
+    const provider = session.provider as StatefulProvider;
+    vi.spyOn(__oauthInternals, 'openExternal').mockImplementation(() => {});
+    const authorizationUrl = new URL('https://example.com/auth?code=xyz');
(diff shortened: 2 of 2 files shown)


gaieges:
Amazing!

steipete:
Closed as superseded by maintainer landing on main in 45dcb65.

I kept the core fix from this PR: the manual OAuth completion URL is now emitted at `warn` so it survives mcporter’s default log threshold. Added focused regression coverage in `tests/oauth-session.test.ts` and a changelog entry crediting @stainlu.

Verification on main:
- `./runner pnpm exec vitest run tests/oauth-session.test.ts tests/runtime-oauth-connect.test.ts tests/runtime-oauth-timeout.test.ts`
- logger-level smoke for default `warn` behavior
- `./runner pnpm check`
- `./runner pnpm build`
- `./runner pnpm test`

PR #147, opened 2026-04-23
fix: respect StdioCommand cwd from config (#146)

### LINKED ISSUE #146: StdioCommand's `cwd` field doesn't work
The `cwd` field isn't used here:

https://github.com/steipete/mcporter/blob/7515dd341282423a1839953aa9980238390afa4f/src/config-normalize.ts#L34-L40

Expected:
If the `cwd` is configured in StdioCommand, the command should be spawned in that `cwd` path.

solomonneas:
## Summary

Fixes #146. The `cwd` field on stdio server entries was silently stripped by
the Zod schema (not declared on `RawEntrySchema`) and then unconditionally
overwritten with the config file directory in `normalizeServerEntry`, so
users could not change the working directory for stdio MCP servers.

## Changes

- `src/config-schema.ts` - add `cwd` to `RawEntrySchema` (optional string).
- `src/config-normalize.ts` - new `resolveCwd()` helper that runs every
  non-empty input through `expandHome` then `path.resolve(baseDir, ...)`:
  - absolute paths are returned normalized
  - a leading `~` is expanded to `$HOME`
  - relative paths resolve against the config file directory
  - omitted or empty `cwd` falls back to the config file directory
- `mcporter.schema.json` - add `cwd` so IDE autocomplete surfaces it.
- `docs/config.md` - add the field to the server definition table.
- `tests/config-normalize.test.ts` - new case covering absolute, relative,
  `~/subdir`, `~`, `''`, and omitted `cwd`.

## Verification

Linux (Ubuntu, Node 22):
- `pnpm check` - clean
- `pnpm test` - 469 passing, 3 skipped

Windows 11 (Node 25, pnpm 10.32):
- `pnpm check` - clean
- `pnpm test` - 454 passing, 18 skipped (pre-existing platform-gated skips)

End-to-end smoke via `node dist/cli.js --config ... config list --json`
confirmed every resolution path lands where expected, and `config doctor`
reports "Config looks good." - i.e. the `path.isAbsolute(cwd)` invariant
used there is preserved.

## Notes

- The Windows run caught a real bug in my first pass: `expandHome` uses `/`
  as its separator, so `~/sub` produced `C:\Users\you/sub`. Normalizing
  through `path.resolve` fixes that without touching `expandHome` (which is
  shared with other callers).
- Intentionally did **not** regenerate `mcporter.schema.json` via
  `pnpm generate:schema`. On current `main`, the generator emits
  reformatted `"required": [...]` arrays unrelated to this change;
  keeping the diff surgical. Happy to reformat the schema in a follow-up
  if you'd prefer.
- `cwd` now resolves relative paths against the config dir. `tokenCacheDir`
  only expands `~` and leaves the rest alone - different semantics by
  field, flagging in case you'd like them aligned in a follow-up.


steipete:
Thanks for tracking this down. I landed the equivalent fix on main in 7ffbd52 with cwd schema/docs coverage and regression tests, and credited this in CHANGELOG.md. Closing this PR as superseded by main.

PR #148, opened 2026-04-23
fix: resolve generated CLI stdio cwd against script location (#56)

### LINKED ISSUE #56: Generated CLI uses relative path for MCP server, fails when run from different directory
## Summary

CLIs generated by mcporter use a relative path (`dist/index.js`) in the `embeddedServer.command.args` configuration. This path is resolved against the current working directory (CWD) at runtime, not relative to the CLI script's location. This causes `MODULE_NOT_FOUND` errors when the CLI is invoked from any directory other than the MCP server's root.

## Environment

- Node.js: v24.11.0
- mcporter: latest (installed via npx)
- OS: macOS

## Steps to Reproduce

1. Create an MCP server project:
```bash
mkdir my-mcp-server && cd my-mcp-server
npm init -y
# ... set up MCP server with dist/index.js as entry point
```

2. Generate CLI using mcporter:
```bash
npx mcporter generate-cli
```

3. The generated `dist/my-cli.cjs` contains:
```javascript
const embeddedServer = {
  "name": "my-cli",
  "command": {
    "kind": "stdio",
    "command": "node",
    "args": ["dist/index.js"]  // <-- relative path
  },
  // ...
};
```

4. Run the CLI from a **different directory**:
```bash
cd /some/other/directory
node /path/to/my-mcp-server/dist/my-cli.cjs some-command --some-arg value
```

## Expected Behavior

The CLI should successfully spawn the MCP server regardless of the current working directory.

## Actual Behavior

```
[mcporter] stderr from node
node:internal/modules/cjs/loader:1423
  throw err;
  ^

Error: Cannot find module '/some/other/directory/dist/index.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1420:15)
    ...
    code: 'MODULE_NOT_FOUND',
    requireStack: []
}

MCP error -32000: Connection closed
```

Node.js resolves `dist/index.js` relative to CWD (`/some/other/directory/`) instead of the CLI script's location.

## Workaround

Manually edit the generated `dist/my-cli.cjs` to resolve the path relative to `__dirname`:

```javascript
// Add before embeddedServer definition:
const __mcpIndexPath = require('path').join(__dirname, 'index.js');

const embeddedServer = {
  "name": "my-cli",
  "command": {
    "kind": "stdio",
    "command": "node",
    "args": [__mcpIndexPath]  // <-- use resolved absolute path
  },
  // ...
};
```

## Suggested Fix

The CLI generator should emit code that resolves the MCP server path relative to the script's location using `__dirname` (for CJS) or `import.meta.url` (for ESM), rather than using a hardcoded relative path.

For CJS bundles:
```javascript
const __mcpServerPath = require('path').join(__dirname, 'index.js');
```

For ESM:
```javascript
import { fileURLToPath } from 'url';
import { dirname, join } from 'path';
const __dirname = dirname(fileURLToPath(import.meta.url));
const __mcpServerPath = join(__dirname, 'index.js');
```

## Impact

This bug prevents using mcporter-generated CLIs in scenarios where:
- The CLI is invoked from a different working directory
- The CLI is globally linked via `npm link`
- The CLI is called by another tool/script running in a different directory
- Shell aliases or scripts invoke the CLI with absolute paths

solomonneas:
Closes #56.

Generated CLIs embed the stdio `cwd` as the generation-time
`process.cwd()` (or omit it entirely for `--command` adhoc inline
definitions), so relative args like `dist/index.js` only resolve when
the CLI is invoked from the exact directory where it was generated.
That breaks `npm link`, global installs, moved bundles, and any shell
with a different cwd - the scenario reported in the issue.

## Fix

- `generate-cli.ts` pre-resolves the final bundle / compile target
  before calling `writeTemplate`, so the template knows where the
  executed artifact will ultimately live (not just the intermediate
  `.ts` template path).
- `template.ts` computes the effective stdio cwd at generation time as
  a path relative to that final artifact's directory.
- The generated CLI emits two constants:
  - `__mcpScriptDir`, resolved from `fileURLToPath(import.meta.url)` at
    runtime. Rolldown's CJS output rewrites this to
    `require('url').pathToFileURL(__filename).href`; Bun's ESM output
    uses it natively.
  - `__mcpRelativeStdioCwd`, the captured relative path (or `null`).
- `normalizeEmbeddedServer` in the generated CLI resolves the relative
  cwd against `__mcpScriptDir` at spawn time.

## Behavior change for adhoc \`--command\`

Before: `--command "node dist/index.js"` embedded no `cwd`, so the
spawn inherited the caller's shell cwd. That's what this issue is
about.
After: the generation-time `process.cwd()` is used as a default and
rewritten relative to the CLI's location, so relative args resolve
portably. In the rare case where a tool genuinely wanted to see the
caller's cwd, users can still set `cwd` explicitly in their config.

## Tests

\`pnpm check\` + \`pnpm test\` clean on Linux (476 passed, 3 skipped) and
Windows 11 (17 passed in the two touched files; existing bun-gated
tests skip as before).

- Unit tests for \`computeRelativeStdioCwd\` cover explicit cwd, adhoc
  no-cwd fallback, HTTP servers, missing output path, same-dir (\".\"),
  and relative cwd inputs.
- \`--bundle\` integration test generates a CLI, invokes the bundle from
  an unrelated cwd, and asserts the spawned server is still found.
  Also exercises a symlinked invocation (skipped on Windows) to cover
  \`npm link\` / global bin shims.
- \`--compile\` integration test gated by \`ensureBunSupport\` covers the
  Bun binary path.
- Windows cross-drive test writes the bundle target to \`D:\\\` while the
  project lives on \`C:\\\` and asserts the \`path.isAbsolute\` guard fires
  so the template emits \`__mcpRelativeStdioCwd = null\`. Gated on
  platform + \`D:\\\` accessibility so it degrades on runners without an
  alt drive.


steipete:
Closed as superseded by the maintainer landing on main in 7691ba8.

I kept the core fix from this PR: generated CLIs now compute a relocation-safe stdio cwd from the final artifact location and resolve it from the generated script at runtime. I also added a regression that reproduces #56 by invoking a generated bundle from an unrelated cwd, plus a path-canonicalization guard for macOS `/var` vs `/private/var`.

Verification on main:
- `./runner pnpm exec vitest run tests/template-cwd.test.ts tests/cli-generate-cli.integration.test.ts`
- manual generated-bundle relocation smoke
- `./runner pnpm check`
- `./runner pnpm build`
- `./runner pnpm test`

Thanks @solomonneas; changelog credit is preserved under PR #148 / issue #56.

PR #149, opened 2026-04-23
fix: quote emit-ts member names for non-identifier tool names

### LINKED ISSUE #30: emit-ts: dashes in names are not handled properly
Specific example: [Notion uses dashes in their tool names](https://developers.notion.com/docs/mcp-supported-tools).

Full notion tools list: https://gist.github.com/jakub300/7e5c07ad68f32801c0c07a9f9595718b

This results in methods and types that are not valid TS syntax as mcporter does not appear to be doing any processing to ensure method names are valid.

solomonneas:
Closes #30.

## Problem

MCP tool names are free-form strings, and servers like Notion (`API-post-page`, `API-patch-block`, ...) expose tools whose names are valid MCP identifiers but invalid JavaScript/TypeScript identifiers. The `emit-ts` command embedded these raw names as bare identifiers in both the generated interface member and the client object-literal method shorthand, which produced TypeScript that fails to parse:

```ts
export interface NotionTools {
  API-post-page(parent: string): Promise<Page>;   // syntax error
}

const client: NotionTools = {
  async API-post-page(params: ...) { ... }        // syntax error
};
```

## Fix

`src/cli/emit-ts-templates.ts` now checks whether the tool name is a bare JS/TS identifier (`/^[A-Za-z_\$][A-Za-z0-9_\$]*\$/`). When it is not, the name is JSON-stringified so the emitted output uses a quoted member name and a bracket-string index access:

```ts
export interface NotionTools {
  "API-post-page"(parent: string): Promise<Page>;
}

const client: NotionTools = {
  async "API-post-page"(params: Parameters<NotionTools["API-post-page"]>[0]) { ... }
};
```

Identifier-safe tool names keep the previous bare-name output, so existing servers see no change. The proxy method name (`entry.methodName`, derived via `toProxyMethodName`) was already sanitized and is untouched.

## Tests

Four new cases in `tests/emit-ts.test.ts`, including one that parses the emitted output with the TypeScript compiler API and asserts zero parse diagnostics. Full suite: 475 passed, 3 skipped. Verified on Linux and Windows 11.

## Scope note

This patch fixes method/type names specifically, which is what the issue reports. Parameter names (schema property names) are emitted verbatim as well, so a tool with a dashed input property would still need quoting. No Notion tool in the referenced list has dashed property names, and handling that cleanly requires a more invasive signature shape, so it is left for a follow-up if someone hits it in the wild.

### DIFF
--- a/src/cli/emit-ts-templates.ts
+++ b/src/cli/emit-ts-templates.ts
@@ -32,7 +32,7 @@ export function renderTypesModule(input: EmitTypesTemplateInput): string {
   lines.push(`export interface ${input.interfaceName} {`);
   input.docs.forEach((entry, index) => {
     lines.push(...renderDocComment(entry.doc.docLines, '  '));
-    const methodSignature = toInterfaceSignature(entry.doc.tsSignature, { wrapInPromise: true });
+    const methodSignature = toInterfaceSignature(entry.doc.tsSignature, entry.toolName, { wrapInPromise: true });
     lines.push(`  ${methodSignature}`);
     if (entry.doc.optionalSummary) {
       lines.push(`  // ${entry.doc.optionalSummary.replace(/^\/\//, '').trim()}`);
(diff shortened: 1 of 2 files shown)


steipete:
Thanks. I landed the equivalent emit-ts quoting fix on main in 6e597cf, including TypeScript parse regression coverage and changelog credit. Closing this PR as superseded by main.

PR #150, opened 2026-04-23
fix: accept key:=value as an alias for key=value so values are not silently dropped

### LINKED ISSUE #100: Numeric parameter syntax (:=) silently drops value for MCP tools with Zod coercion
## Bug Description

When using the `:=` syntax to pass numeric parameters via the mcporter CLI, the value is silently dropped and not sent to the MCP server. The string syntax (`=`) works correctly because Zod coerces the string to a number.

## Steps to Reproduce

```bash
# This FAILS — price is silently dropped, server receives no price field
mcporter call schwab placeOrder session=NORMAL duration=DAY orderType=LIMIT price:=5.20 orderStrategyType=SINGLE \
  'orderLegCollection=[{"instruction":"BUY_TO_CLOSE","quantity":1,"instrument":{"assetType":"OPTION","symbol":"TSLA  260417P00350000"}}'

# This WORKS — price arrives as string "5.20", Zod coerces to number
mcporter call schwab placeOrder session=NORMAL duration=DAY orderType=LIMIT price=5.20 orderStrategyType=SINGLE \
  'orderLegCollection=[{"instruction":"BUY_TO_CLOSE","quantity":1,"instrument":{"assetType":"OPTION","symbol":"TSLA  260417P00350000"}}'
```

## Expected Behavior

`price:=5.20` should pass `price` as the number `5.20` to the MCP tool.

## Actual Behavior

The `price` field is silently dropped. The MCP server never receives it. Confirmed by inspecting the server-side logs — the order body JSON has no `price` field at all.

The server (Schwab MCP via Cloudflare Workers) then sends the request to the upstream API with price=0, which returns `"Limit price cannot be zero for limit orders."`.

## Environment

- mcporter: 0.7.3
- Node: v25.6.0
- macOS (arm64)
- MCP server: custom Schwab trading API (Zod schema expects `price?: number`)

## Workaround

Use `price=5.20` (string syntax) instead of `price:=5.20` (numeric syntax). Zod coerces the string to a number correctly.

## Notes

This may affect any MCP tool with optional numeric parameters validated by Zod. The `:=` syntax is documented as the way to pass typed (non-string) values, so users would naturally reach for it when passing numbers.

solomonneas:
Closes #100.

When a user passes `price:=5.20` (the syntax `httpie` users reach for to send a typed value), mcporter currently calls `indexOf('=')` first in `parseKeyValueToken`, which slices the key as `"price:"` with a trailing colon and stores the value under that malformed key. The MCP server's Zod schema has no `price:` field, so the entry is ignored and the tool behaves as if `price` were never provided.

This change detects a `:` immediately before the first `=` and excludes it from the key, so `key:=value` is parsed the same way as `key=value`. It does not introduce any httpie-style typed-value semantics - the string-coercion candidate registration (used for IDs like Slack `thread_ts=...`) still applies, and `key=value`, `key:value`, `key: value` are unaffected. `:` inside the value is also unaffected because only the character immediately preceding `=` is checked.

### Changes
- `src/cli/call-argument-values.ts` - one-line parser fix: trim a trailing `:` from the key when it sits immediately before `=`.
- `tests/call-arguments.test.ts` - four new cases: the reported Schwab repro, zero/integer coercion, schema string-coercion candidate parity, and a non-regression for `:=` appearing inside a value (`expr=value:=x` stays as `value:=x`).
- `docs/call-syntax.md` - document `:=` as a compatibility alias and call out that it does not enable typed-value semantics.

### Verification
- `pnpm check` - oxfmt, oxlint type-aware, tsgo all clean.
- `pnpm test` - 472 passed, 3 skipped, 0 failed.
- Built and smoke-tested `parseCallArguments` with the exact tokens from the issue repro (`price:=5.20`, `quantity:=0`, `limit:=10`); args object now contains the expected keys.

### DIFF
--- a/docs/call-syntax.md
+++ b/docs/call-syntax.md
@@ -67,7 +67,7 @@ Key details:
 
 ## Flag-Based Syntax Details
 
-- `key=value`, `key:value`, and `key: value` all map to the same named-argument handling, so you can type whichever feels most natural for your shell.
+- `key=value`, `key:value`, `key: value`, and `key:=value` all map to the same named-argument handling, so you can type whichever feels most natural for your shell. The `:=` form is accepted as a compatibility alias for `=`; it does not enable any typed-value semantics like httpie's `:=`.
 - By default, arguments keep the same validation pipeline as the function-call syntax—enums, numbers, and booleans are coerced automatically, and missing required fields raise errors.
 - Numeric-looking `key=value` arguments are restored to their original string spelling when the tool schema declares that parameter as a string, which keeps timestamp-like IDs such as Slack `thread_ts=1234567890.123456` intact.
 - `--raw-strings` disables numeric coercion for flag-style and positional values so IDs/codes stay literal strings (`code=12345` stays `"12345"`).
--- a/src/cli/call-argument-values.ts
+++ b/src/cli/call-argument-values.ts
@@ -9,7 +9,8 @@ export interface ParsedKeyValueToken {
 export function parseKeyValueToken(token: string, nextToken: string | undefined): ParsedKeyValueToken | undefined {
   const eqIndex = token.indexOf('=');
   if (eqIndex !== -1) {
-    const key = token.slice(0, eqIndex);
+    const keyEnd = eqIndex > 0 && token[eqIndex - 1] === ':' ? eqIndex - 1 : eqIndex;
+    const key = token.slice(0, keyEnd);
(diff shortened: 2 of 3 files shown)


steipete:
Thanks. I landed the equivalent key:=value parser fix on main in 6e597cf with regression coverage and changelog credit, so closing this PR as superseded by main.

PR #151, opened 2026-04-24
fix: avoid truncating stdout before forced exit

### LINKED ISSUE #145: force-exit may truncate large stdout when mcporter is called via child_process
Hi, I found a reproducible issue with the current forced-exit behavior and wanted to check whether you'd be open to a small design change before I send a PR.
## Summary
When `mcporter` is invoked from Node via `child_process.exec` / `execFile`, large stdout payloads can get truncated if `MCPORTER_NO_FORCE_EXIT` is not set.
This seems to happen because the CLI currently does a forced `process.exit(0)` very soon after cleanup, which can race with stdout/stderr flushing when the parent process is reading from a pipe.
With `MCPORTER_NO_FORCE_EXIT=1`, the same command returns the full payload.
## Platform notes
I have reproduced this on macOS.
I have not been able to reproduce it on Linux so far.
So at the moment this looks like a macOS-specific pipe/flush timing issue, or at least something that is much easier to trigger there.
## Repro context
In my real-world case, `hyy-bi-mcp` is a private MCP service with a large number of business-facing tools.
So `mcporter list hyy-bi-mcp --schema --output json` produces a fairly large response, around 200 KB.
That large payload size is what makes the truncation easy to observe in practice.
## Repro
I reproduced this with:
```js
const { exec } = require('child_process');
exec(
  'mcporter list hyy-bi-mcp --schema --output json',
  { maxBuffer: 10 * 1024 * 1024 },
  (err, stdout, stderr) => {
    console.log({
      err: err && { message: err.message, code: err.code, signal: err.signal },
      stdoutBytes: Buffer.byteLength(stdout || ''),
      stderrBytes: Buffer.byteLength(stderr || ''),
      stdoutTail: (stdout || '').slice(-120),
    });
  }
);
```
## Observed behavior
On macOS, without MCPORTER_NO_FORCE_EXIT=1:
- stdoutBytes is truncated
- in one repro it stopped exactly at 8192 bytes
- the JSON payload is incomplete / unparsable

On macOS, with MCPORTER_NO_FORCE_EXIT=1:
- the full JSON is returned
- payload parses normally

On Linux:
- I have not reproduced the truncation so far with the same basic scenario

## Suspected cause
The current logic in src/cli.ts forces exit after runtime cleanup to avoid hangs caused by lingering stdio handles / SDK shutdown issues.
That makes sense, but in the child-process case it appears to exit before the stdio pipe is fully flushed to the parent.
So this looks like a tradeoff between:
1. avoiding hangs from leaked handles
2. avoiding truncation of large stdout/stderr payloads

## Proposed design
Would you be open to changing the default force-exit behavior to something like this?
1. after cleanup, set process.exitCode = 0
2. let Node exit naturally if stdio flushes quickly
3. keep a short fallback timer (for example 50ms, maybe configurable) that calls process.exit(0) only if the process still hasn't exited
That would preserve the anti-hang behavior while giving piped stdout/stderr a chance to drain before the process is hard-killed.
Pseudo-shape:
```js 
process.exitCode = 0;
setTimeout(() => {
  process.exit(0);
}, 50);
```
The key idea is:
- do not hard-exit immediately on the next tick
- do keep a short hard-exit fallback so the original hang workaround still exists
## Why I’m asking first
I already reproduced this locally and tested a small implementation along these lines, but I wanted to confirm that this direction matches your intended CLI behavior before opening a PR.
If this approach sounds reasonable, I can send:
- a minimal patch
- a regression test covering child_process.exec/execFile + large JSON output
- no behavior change for the MCPORTER_NO_FORCE_EXIT=1 escape hatch

Would you prefer this grace-period approach, or would you rather solve it another way?

yuhp:
## Summary
- avoid truncating large stdout/stderr when `mcporter` is invoked via `child_process`
- replace the immediate forced exit with a short grace period before the fallback `process.exit(0)`
- add a regression test covering large JSON output under the default force-exit path
## Why
The CLI currently forces `process.exit(0)` immediately after cleanup unless `MCPORTER_NO_FORCE_EXIT=1` is set.
On macOS, this can race with stdio pipe flushing when `mcporter` is called from Node via `child_process.exec` / `execFile`, which can truncate large outputs. In the reported real-world case, `mcporter list ... --schema --output json` produced about 200 KB of JSON from a private MCP service and the parent process sometimes only received the first chunk.
This change keeps the anti-hang workaround in place while giving Node a short chance to flush buffered output naturally before the fallback forced exit runs.
## Issues
Closes #145.
## Changes
- add `FORCE_EXIT_GRACE_MS = 50` in `src/cli.ts`
- change the default forced-exit path to:
  - set `process.exitCode = 0`
  - schedule a short fallback `process.exit(0)` instead of exiting immediately on the next tick
- add `tests/cli-force-exit-flush.integration.test.ts`
  - starts a local stdio MCP server
  - generates a large `--schema --output json` payload
  - verifies the output is not truncated and remains valid JSON when called through `execFile`
## Testing
- `./runner pnpm build`
- `./runner pnpm exec vitest run tests/cli-force-exit-flush.integration.test.ts`
- `./runner pnpm exec vitest run tests/stdio-servers.integration.test.ts tests/cli-list-verbose-e2e.test.ts`
## Notes
- I reproduced the truncation on macOS
- I have not reproduced the same issue on Linux so far
- `./runner pnpm check` could not complete in this environment because `oxfmt` is not available (`sh: oxfmt: command not found`)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -27,6 +27,8 @@ export { handleInspectCli } from './cli/inspect-cli-command.js';
 export { extractListFlags, handleList } from './cli/list-command.js';
 export { resolveCallTimeout } from './cli/timeouts.js';
 
+const FORCE_EXIT_GRACE_MS = 50;
+
 export async function runCli(argv: string[]): Promise<void> {
   const args = [...argv];
   if (args.length === 0) {
@@ -195,7 +197,10 @@ export async function runCli(argv: string[]): Promise<void> {
       } else {
         const scheduleExit = () => {
           if (!disableForceExit || process.env.MCPORTER_FORCE_EXIT === '1') {
-            process.exit(0);
+            process.exitCode = 0;
+            setTimeout(() => {
+              process.exit(0);
+            }, FORCE_EXIT_GRACE_MS);
           }
         };
         setImmediate(scheduleExit);
--- a/tests/cli-force-exit-flush.integration.test.ts
+++ b/tests/cli-force-exit-flush.integration.test.ts
@@ -1,0 +1,135 @@
+import { execFile } from 'node:child_process';
+import fs from 'node:fs/promises';
+import { createRequire } from 'node:module';
+import os from 'node:os';
+import path from 'node:path';
+import process from 'node:process';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+const CLI_ENTRY = fileURLToPath(new URL('../dist/cli.js', import.meta.url));
+const PNPM_COMMAND = process.platform === 'win32' ? 'cmd.exe' : 'pnpm';
+const PNPM_ARGS_PREFIX = process.platform === 'win32' ? ['/d', '/s', '/c', 'pnpm'] : [];
+const testRequire = createRequire(import.meta.url);
+const MCP_SERVER_MODULE = pathToFileURL(testRequire.resolve('@modelcontextprotocol/sdk/server/mcp.js')).href;
+const STDIO_SERVER_MODULE = pathToFileURL(testRequire.resolve('@modelcontextprotocol/sdk/server/stdio.js')).href;
+const ZOD_MODULE = pathToFileURL(path.join(process.cwd(), 'node_modules', 'zod', 'index.js')).href;
+
+function pnpmArgs(args: string[]): string[] {
+  return [...PNPM_ARGS_PREFIX, ...args];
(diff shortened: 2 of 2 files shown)


steipete:
Thanks for the repro and patch direction. I landed the equivalent fix on main in e1a35f8, with an integration test that exercises the built CLI through child_process and verifies large JSON output is not truncated. Credited in CHANGELOG.md; closing this PR as superseded by main.

Issue #117, opened 2026-03-17
Support --header flag for ad-hoc HTTP servers (--http-url)


lxw322:
## Problem

When using `--http-url` to connect to ad-hoc StreamableHTTP MCP servers, there is no way to pass custom HTTP headers (e.g., `Authorization`, `X-API-Key`, `X-Tenant`). Many private/internal MCP servers require authentication headers to accept connections.

Currently, `--header` is **only supported by `mcporter config add`**, not by the ephemeral ad-hoc flags (`--http-url`). The `EphemeralServerSpec` interface lacks a `headers` field, and `extractEphemeralServerFlags()` does not parse `--header`.

This forces users into a two-step workflow:

```bash
# Cannot do this today:
npx mcporter list --http-url https://internal.example.com/mcp --header "Authorization=Bearer sk-xxx"

# Must do this instead:
npx mcporter config add my-service https://internal.example.com/mcp --header "Authorization=Bearer sk-xxx"
npx mcporter list my-service
```

## Proposed Solution

Add `--header KEY=value` (repeatable) support to the ephemeral server flags, consistent with how `--env` already works in ad-hoc mode.

### Changes needed

1. **`EphemeralServerSpec`** (`adhoc-server.ts`): add `headers?: Record<string, string>`
2. **`extractEphemeralServerFlags()`** (`ephemeral-flags.ts`): parse `--header KEY=value`, same as `--env`
3. **`resolveEphemeralServer()`** (`adhoc-server.ts`): pass `spec.headers` into the `HttpCommand.headers` field (merged with `ensureHttpAcceptHeader`)
4. **`persistEphemeralServer()`**: include headers in the persisted entry when `--persist` is used

### Example usage

```bash
# Quick exploration with auth
npx mcporter list --http-url https://api.example.com/mcp \
  --header "Authorization=Bearer sk-xxx" \
  --header "X-Tenant=biz-unit-01"

# Call with headers
npx mcporter call --http-url https://api.example.com/mcp \
  --header "Authorization=Bearer sk-xxx" \
  my-tool param1=value1

# Persist for future use
npx mcporter list --http-url https://api.example.com/mcp \
  --header 'Authorization=$env:MY_TOKEN' \
  --name my-service \
  --persist config/mcporter.json
```

### Environment variable placeholders

Should support the same `$env:VAR` / `${VAR}` / `${VAR:-default}` syntax already used in config file headers, so sensitive values don't leak into shell history:

```bash
[...]

PR #152, opened 2026-04-24
feat: support custom HTTP headers via ENVHEADER environment variable

dalinna:
Add support for injecting custom HTTP headers through the ENVHEADER environment variable using JSON format. This enables distributed tracing and other observability features when mcporter is called from external systems like OpenClaw.

The ENVHEADER variable accepts a JSON object where each key-value pair represents a header name and value:

  ENVHEADER='{"traceparent":"00-xxx-xxx-01","x-request-id":"abc123"}'

Features:
- Case-insensitive header name collision detection
- Silent failure on invalid JSON to maintain backward compatibility
- Only string values are accepted as header values
- Environment headers don't override existing configured headers

This change enables seamless integration with observability systems without requiring users to modify mcporter configuration files or pass headers explicitly on every call.

### DIFF
--- a/src/runtime/transport.ts
+++ b/src/runtime/transport.ts
@@ -117,8 +117,14 @@ function createHttpTransportOptions(
         // Only add headers that don't already exist (case-insensitive check)
         const existingHeadersLower = new Set(Object.keys(finalHeaders).map(k => k.toLowerCase()));
         for (const [key, value] of Object.entries(envHeaders)) {
+          // Skip Authorization header if OAuth is being established to prevent conflicts
+          if (shouldEstablishOAuth && key.toLowerCase() === 'authorization') {
+            continue;
+          }
           if (!existingHeadersLower.has(key.toLowerCase()) && typeof value === 'string') {
             finalHeaders[key] = value;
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the PR. I landed the explicit ad-hoc header path on main in 3f4f8dc, which covers the user-facing gap from #117 without adding a hidden global header side channel:

```bash
mcporter list --http-url https://api.example.com/mcp \
  --header 'Authorization=$env:MY_TOKEN' \
  --header 'X-Tenant=biz-unit-01'
```

I am not landing `ENVHEADER` as-is because it silently mutates every HTTP transport from one undocumented environment variable, ignores invalid payloads, and sits outside the existing config/ad-hoc/header resolver model. The explicit `--header` path is easier to audit, works for `list`/`call`/`auth`, and persists cleanly when users pass `--persist`.

Verification for the landed fix:
- focused ad-hoc/runtime shards
- parser/resolve smoke with `$env:MY_TOKEN` placeholders
- built CLI help smoke for `list`, `call`, and `auth`
- full `check`, `build`, and `test`

Appreciate the contribution; this helped point at the same missing header surface.

PR #154, opened 2026-04-28
fix: return non-zero exit code on MCP tool errors

### LINKED ISSUE #153: CLI `call` commands return exit code 0 on MCP tool errors
When using the `mcporter call` CLI to execute a tool on an MCP server, if the tool execution fails (e.g., due to an 'Unknown resource' or other MCP protocol error), the CLI prints the error but exits with exit code 0. This breaks LLM agent orchestrators that rely on the exit code to determine if the tool call succeeded (e.g., setting `isError: false`). It should exit with a non-zero exit code (e.g. 1) on tool execution failure.

jlapenna:
Fixes #153

When a tool returns `isError: true`, the CLI should exit with a non-zero exit code so that orchestrators (like LLM agents) can detect the failure. This also patches an issue in `cli.ts` where `process.exit(0)` was overriding `process.exitCode`.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -190,12 +190,12 @@ export async function runCli(argv: string[]): Promise<void> {
       if (DEBUG_HANG) {
         dumpActiveHandles('after terminateChildProcesses');
         if (!disableForceExit || process.env.MCPORTER_FORCE_EXIT === '1') {
-          process.exit(0);
+          process.exit(process.exitCode ?? 0);
         }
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for the fix. I landed the equivalent behavior on main in e1a35f8, including the exit-code preservation path and regression coverage. I also credited this in CHANGELOG.md for the next release, so closing this PR as superseded by main.

PR #161, opened 2026-05-08
fix: resolve env placeholders across config fields

### LINKED ISSUE #157: Resolve `${VAR}` env-var placeholders in all string-valued `mcporter.json` fields
### Problem

mcporter already resolves `${VAR}` and `${VAR:-fallback}` placeholders inside two fields: `headers` and `env` (per [`config-schema.ts:65-72`](https://github.com/openclaw/mcporter/blob/main/src/config-schema.ts#L65)). For every other env-driven configuration value, the project ships a dedicated `*Env` companion field — e.g. `bearerToken` / `bearerTokenEnv`, `oauthClientSecret` / `oauthClientSecretEnv`. A new env-driven field requires:

1. A new `*Env` companion in the schema.
2. A new resolver next to existing ones.
3. Documentation explaining which fields support env redirection.

This pattern doesn't scale and the asymmetry creates real gaps. Most visible today: there is no `oauthClientIdEnv` — only `oauthClientId` (literal). Providers using DCR with per-instance client_ids therefore can't supply the value via env without baking per-deployment identifiers into published bundles. The same pattern recurs whenever a new config field needs env-redirection.

### Proposal

Make `${VAR}` and `${VAR:-fallback}` placeholders work uniformly in all string-valued config fields, applied at the same resolution point as the existing `headers`/`env` substitution. Concretely:

1. Extract the existing placeholder resolver from the `headers` / `env` paths into a shared helper.
2. Apply it to all string-valued fields in `RawEntrySchema` (or a documented opt-in subset such as `baseUrl`, `oauthClientId`, `oauthClientSecret`, `tokenCacheDir`, `clientName`, `oauthRedirectUrl`, `oauthScope`).
3. Deprecate `*Env` companion fields in favor of placeholder usage. New code paths consume the resolved value directly; old `*Env` fields keep working until removed in a major version.

**Example** — today:

```jsonc
{
  "mcpServers": {
    "linear": {
      "baseUrl": "https://mcp.linear.app/mcp",
      "auth": "oauth",
      "oauthClientSecretEnv": "LINEAR_MCP_CLIENT_SECRET"
      // no equivalent for client_id — must be literal
    }
  }
}
```

After this change:

```jsonc
{
  "mcpServers": {
    "linear": {
      "baseUrl": "https://mcp.linear.app/mcp",
      "auth": "oauth",
      "oauthClientId": "${LINEAR_MCP_CLIENT_ID}",
      "oauthClientSecret": "${LINEAR_MCP_CLIENT_SECRET:-}"
    }
  }
}
```

### Why generalize instead of adding `oauthClientIdEnv`

Adding one more `*Env` companion fixes the immediate gap but reproduces the underlying schema-bloat problem. A single uniform mechanism scales: any future field that integrators need to inject from env works without further schema changes. It also matches what users already do with `headers` / `env` and aligns with conventional shell/template behavior.

### Failure-mode considerations

- **Bare `${VAR}` with the env unset.** Today's `headers`/`env` resolver expands to empty silently. For sensitive fields (client_id, client_secret) silent empty-string is unhelpful. Consider failing loudly when a placeholder without `:-fallback` resolves to empty — matches operator intent. Existing `${VAR:-fallback}` syntax stays the explicit "I'm OK with absence" opt-in.
- **Resolution timing.** `headers`/`env` resolve per-request; OAuth fields resolve once at OAuth-flow init. Both are fine — the resolver only needs to run when the value is read; existing call-sites already gate that.
- **Backward compatibility.** Existing literal values that happen to contain `$` (rare, but possible) need a documented escape (e.g. `$$`) or a per-field opt-in flag. The `headers`/`env` substitution today doesn't define an escape — worth resolving as part of the same change.

### Why now

DCR-based MCP servers like Linear's `mcp.linear.app` issue per-installation client_ids that integrators don't want to commit to published bundles. The current workaround (vault-seed the `clientInfo` externally — see #156) couples consumers to mcporter's internal vault format. Adding `oauthClientIdEnv` would solve this one case but the next field with the same need triggers the same conversation again. Generalizing the placeholder mechanism closes the entire class.

zxyasfas:
## Summary

Resolves `${VAR}` and `${VAR:-fallback}` placeholders uniformly across string-valued `mcporter.json` server fields instead of limiting placeholder support to headers and env overrides.

## Changes

- Reuses the existing environment placeholder resolver during config normalization.
- Supports fallback placeholders inside interpolated strings, e.g. `https://${HOST:-example.com}/mcp`.
- Applies resolution recursively across nested string values such as OAuth fields, command args, cwd, env records, and OAuth command args.
- Adds coverage for successful cross-field substitution and missing required placeholder errors.

Fixes #157.

## Validation

- Not run locally; the repository checkout/download timed out in this environment, so the patch was prepared against the targeted source files via the GitHub API.

### DIFF
--- a/src/config-normalize.ts
+++ b/src/config-normalize.ts
@@ -1,7 +1,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import type { CommandSpec, RawEntry, ServerDefinition, ServerLoggingOptions, ServerSource } from './config-schema.js';
-import { expandHome } from './env.js';
+import { expandHome, resolveEnvPlaceholders } from './env.js';
 import { resolveLifecycle } from './lifecycle.js';
 
 export function normalizeServerEntry(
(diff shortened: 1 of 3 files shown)


steipete:
Thanks @zxyasfas. I landed the fix on main in ea91086 and credited the PR in the changelog.

I kept the same core behavior from this PR: `` / `fallback` now works for string-valued config fields, including `oauthClientId` and the other static OAuth fields.

One tweak from the branch: I left `headers`, `env`, bearer-token placeholders, and `*Env` variable-name fields unresolved at config-normalization time. Those continue to resolve at runtime, which avoids `mcporter config get/list` materializing secrets.

Proof: `./runner pnpm check`, `./runner pnpm test` (543 passed, 3 skipped), live temp-config CLI smoke, and green main CI: https://github.com/openclaw/mcporter/actions/runs/25600859609

PR #172, opened 2026-05-14
feat: expose daemon-managed MCP servers via serve

zm2231:
## Summary

Adds `mcporter serve`, a bridge that exposes daemon-managed keep-alive servers as a single MCP endpoint.

The mcporter daemon already keeps keep-alive servers warm between calls, but MCP clients such as Claude Code and Codex cannot connect to it directly because it speaks mcporter's private Unix-socket protocol. Without a bridge, each client session still launches its own copy of every configured MCP server.

`mcporter serve` sits between the daemon-backed runtime and any MCP client. Clients register one MCP server, while mcporter handles tool discovery, namespacing, routing, and lifecycle reuse through the daemon.

## What's Included

- `src/serve.ts`
  - MCP SDK `McpServer` bridge.
  - Accepts a `Runtime` surface (`listTools` / `callTool`) rather than a raw `DaemonClient`.
  - Exposes selected keep-alive tools as `server__tool` to avoid cross-server name collisions.
  - Escapes bridged server/tool name parts so configured names containing `__` cannot collide.
  - Supports stdio and Streamable HTTP transports.

- `src/cli/serve-command.ts`
  - Adds `mcporter serve`.
  - Wires `createRuntime` + `DaemonClient` + `createKeepAliveRuntime`.
  - Defaults to stdio.
  - Supports `--http <port>` on `127.0.0.1` by default.
  - Supports `--host <host>` for explicit non-local binding.
  - Supports `--servers a,b,c` to restrict exposed keep-alive servers.

## Review Fixes

- Rejects empty equals-form HTTP ports such as `--http=` instead of treating them as port `0`.
- Adds bridge coverage with a real `createKeepAliveRuntime` wrapper to prove served keep-alive tool listing/calls route through the daemon client and trigger restart/close behavior on fatal daemon errors.
- Escapes bridged tool-name components to prevent collisions such as `alpha` + `beta__ping` and `alpha__beta` + `ping`.

## Tests

Added coverage in:
- `tests/serve.test.ts`
- `tests/cli-serve-command.test.ts`
- `tests/cli-serve-runtime.test.ts`

Validated with:
- `./runner pnpm check`
- `./runner pnpm docs:list`
- `./runner pnpm exec tsx src/cli.ts serve --help`
- `./runner pnpm test`

Result:
- 116 test files passed, 1 skipped
- 571 tests passed, 3 skipped

## Forward Compatibility

If `createManagedRuntime()` lands later, only `serve-command.ts` needs to swap its manual runtime wiring for that helper. `serve.ts` should not need changes because it already depends on the runtime interface rather than daemon construction details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ npx mcporter call --stdio "bun run ./local-server.ts" --name local-tools
 - Stop it anytime with `mcporter daemon stop`, pre-warm with `mcporter daemon start`, or bounce it via `mcporter daemon restart` after tweaking configs/env.
 - All other servers stay ephemeral; add `"lifecycle": "keep-alive"` to a server entry (or set `MCPORTER_KEEPALIVE=name`) when you want the daemon to manage it. You can also set `"lifecycle": "ephemeral"` (or `MCPORTER_DISABLE_KEEPALIVE=name`) to opt out.
 - The daemon only manages named servers that come from your config/imports. Ad-hoc STDIO/HTTP targets invoked via `--stdio …`, `--http-url …`, or inline function-call syntax remain per-process today; persist them into `config/mcporter.json` (or use `--persist`) if you need them to participate in the shared daemon.
+- `mcporter serve --stdio` exposes every daemon-managed keep-alive server as one MCP stdio bridge for clients such as Claude Code or Codex. Register it once, then call namespaced tools like `chrome-devtools__list_pages`; add `--servers a,b` to limit the bridge or `--http <port>` to serve Streamable HTTP on localhost at `/mcp`.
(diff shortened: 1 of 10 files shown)


steipete:
Landed on `main` in 8d962fb.

I could not update this PR branch directly because `maintainerCanModify` is false, so I recreated the branch in the maintainer repo, added the required changelog entry and one fix to keep bridged names readable (`memory__create_entities` instead of percent-escaping ordinary underscores), then landed that tested branch.

Proof run locally before landing:
- `./runner pnpm check`
- `./runner pnpm test`
- `./runner pnpm docs:list`
- manual `mcporter serve --http 0` e2e against a keep-alive stdio memory fixture, including list/call through the bridge

Thanks @zm2231. For future PRs, enabling "Allow edits by maintainers" lets us push small fixups directly to the PR branch.

PR #176, opened 2026-05-16
feat(health): add mcporter health for at-a-glance server status

mvanhorn:
## Summary

- Adds `mcporter health`. Pings every configured MCP server in parallel (via `Promise.allSettled`), reports per-server status, initialize latency, tool count, and OAuth token state in one table. Exits non-zero when any server is not ok, so it works as a CI/uptime monitor.
- Flags: `--server <name>` to check one, `--timeout <seconds>` per-server (default 10), `--json` for machine-readable output.
- Classification handles the common failure modes:
  - `auth_required` for 401 / forbidden / token errors
  - `unreachable` for ECONNREFUSED / ETIMEDOUT / ENOTFOUND
  - `error` for anything else (with a 200-char preview from the error message)
  - `ok` with measured latency and tool count when healthy
- Subcommand follows the existing `list-command.ts` / `daemon-command.ts` pattern: commander registration, reuses `output-utils.ts` / `json-output.ts` / `terminal.ts` for formatting. No new runtime deps.

## Why this matters

`mcporter list` shows the servers configured for the current workspace and their tool counts, but not whether those servers are actually reachable or whether their cached OAuth tokens are still valid. At 10+ MCPs configured — the audience that exists at 4.4K stars — there is no way to check "is everything ok before I start a run" short of calling each tool manually.

This is operational quality-of-life that compounds at scale:

```bash
mcporter health                         # check all configured servers
mcporter health --server linear         # check one
mcporter health --json                  # CI / uptime monitor friendly
mcporter health --timeout 5             # tighter per-server budget
```

Parallel execution means one slow or dead server does not block the others.

## Demo

Simulated demo:

mcporter health demo

The demo shows a 5-server config: 3 healthy (linear / context7 / vercel with latency + tool counts + OAuth state), 1 with an expired token (supabase → `auth_required`), 1 unreachable (internal-api on ECONNREFUSED). Exit code 1 because not everything is ok.

## Testing

- `corepack pnpm typecheck`
- `corepack pnpm lint` (oxlint clean, oxfmt clean)
- `corepack pnpm test` — 648 tests pass; new `tests/cli-health-command.test.ts` covers:
  - all-servers-reachable path returns 0, table has one row per server
  - one server returning 401 → row classified as `auth_required`, exit code 1
  - timeout → row classified as `unreachable` with error preview
  - `--json` produces a valid JSON array conforming to the HealthRow shape
  - `--server <name>` filter checks only the named server
  - `--timeout 1` enforces per-server timeout via stubbed slow transport

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## [0.11.2] - Unreleased
 
-- Nothing yet.
+### Added
+
+- Add `mcporter health` for at-a-glance server status, latency, tool count, OAuth state, and JSON/quiet output.
 
 ## [0.11.1] - 2026-05-14
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ MCPorter helps you lean into the "code execution" workflows highlighted in Anthr
 
 - **Zero-config discovery.** `createRuntime()` merges your home config (`~/.mcporter/mcporter.json[c]`, or `$XDG_CONFIG_HOME/mcporter/mcporter.json[c]` when set) first, then `config/mcporter.json`, plus Cursor/Claude/Codex/Windsurf/OpenCode/VS Code imports, expands `${ENV}` placeholders, and pools connections so you can reuse transports across multiple calls.
 - **One-command CLI generation.** `mcporter generate-cli` turns any MCP server definition into a ready-to-run CLI, with optional bundling/compilation and metadata for easy regeneration.
+- **Health checks.** `mcporter health` pings every configured server and reports status, latency, tool count, and OAuth state in one table.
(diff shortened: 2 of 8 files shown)


steipete:
Thanks for opening this. I reviewed the shape and found we already had the same status concept in `mcporter list`: multi-server `list` already probes servers and reports auth/offline/http/error counts, while `daemon status` is the daemon-specific status surface.

I landed #183 / de7c811 instead of adding a new top-level `health` command. It folds the missing health-check behavior into `mcporter list` with:

- `mcporter list --quiet`
- `mcporter list --exit-code`
- `mcporter list <server> --status`
- JSON/status output that stays parseable even when failing stdio servers write stderr

That keeps the CLI surface smaller and reuses the existing list classification path. Closing this as superseded by #183. Thanks again for pushing the at-a-glance status idea forward.

Issue #177, opened 2026-05-16
OAuth callback server not reachable when run via process managers that use process groups (e.g. OpenClaw exec)


qyc-naro:
### Description
When `mcporter auth <server>` is run through process managers that spawn commands in a process group (such as OpenClaw's exec tool), the OAuth callback HTTP server either never starts or gets killed before the browser redirect can complete.

### Steps to Reproduce
1. Configure an OAuth-based MCP server (e.g., Naro, Composio)
2. Run `mcporter auth naro` through OpenClaw's exec tool (or any process manager that uses process groups)
3. The OAuth URL is printed
4. Open the URL in browser, sign in, approve scopes
5. Browser redirect to `127.0.0.1:<port>/callback` fails — connection refused

### Expected Behavior
- mcporter starts a local HTTP listener on the callback port
- It waits (with timeout) for the OAuth redirect callback
- After receiving the callback, it exchanges the code for tokens and persists them

### Actual Behavior
- The OAuth URL is printed to stdout
- The process exits immediately (when run via exec tool) or the listener is never bound
- `lsof -i :<port>` shows nothing listening
- No tokens are saved

### Root Cause
When run directly from a terminal, `mcporter auth` works correctly:
- The process stays running
- The HTTP server binds to the callback port
- `lsof` confirms the port is listening

When run via OpenClaw's exec tool (which uses process groups), the process tree gets cleaned up when the exec session ends. The mcporter process is killed before it can:
1. Start the HTTP listener, OR
2. Complete the OAuth handshake

Verified with `nohup` workaround — running `nohup mcporter auth composio --reset &` keeps the process alive and the port listening correctly.

### Workaround
Run `mcporter auth <server>` directly from a terminal (not through OpenClaw exec), or use:
```bash
nohup mcporter auth <server> > /tmp/mcporter_auth.txt 2>&1 &
```
Then open the printed URL in your browser and complete the OAuth flow.

### Environment
- mcporter version: 0.11.1
- Node.js: v26.0.0
- OS: macOS (Darwin 25.4.0, arm64)
- OpenClaw: 2026.5.12

qyc-naro:
## Update: Confirmed root cause

Tested extensively. The mcporter code itself is correct — `http.createServer()` and `server.listen()` work fine when the process stays alive.

**Evidence:**
- Running `mcporter auth composio --reset` directly from terminal: process stays running, port IS listening (`lsof` confirms)
[...]

PR #178, opened 2026-05-16
docs: clarify headless OAuth process lifetime

asim48-ctrl:
## Summary
- Document that `mcporter auth --no-browser` still needs the CLI process to stay alive until the loopback callback completes
- Call out process managers that capture stdout and then clean up the process group, leaving no callback listener
- Add persistent-terminal/tmux/nohup guidance in the README, ad-hoc auth docs, and known issues

Refs #177.

## Validation
- `git diff --check`
- Not run: full docs/test gate (docs-only change; local Bun/pnpm validation avoided to keep this lightweight)


steipete:
Verified locally before merge:
- `git diff --check main...HEAD`
- `./runner pnpm format:check`

This is docs-only, so I skipped Codex Auto Review and the full test suite.

Issue #179, opened 2026-05-17
MCPorter not using OAuth credentials


jaigew:
Using mcporter v. 0.11.1, after adding the [courtlistener MCP server](https://wiki.free.law/c/courtlistener/help/api/mcp/model-context-protocol-mcp-server-for-agentic-access) through mcporter and authenticating it with  oauth, every call to the MCP server requires reauthentication.  If I reauthenticate, the MCP server responds correctly. 

**Expected result:**
- after authentication, the credentials are reused with subsequent calls.
- For example, issuing "mcporter call 'courtlistener.search' type="o" q="Brown v. Board of Education" num_results=1 --output json" after authentication should simply execute. 

**Actual result:**
- After authentication, every call to the MCP server still requires authentication again.  
- If I reauthenticate through oauth, the calls to the MCP server work correctly.  

Results of mcporter list shows courtlistener is set up and authenticated.  

```
mcporter 0.11.1 — Listing 1 server(s) (per-server timeout: 30s)
- courtlistener (14 tools, 0.9s)
✔ Listed 1 server (1 healthy).

```
~/.mcporter/mcporter.json and ./config/mcporter.json are identical and include auth:"oauth":

```
{
  "mcpServers": {
    "courtlistener": {
      "baseUrl": "https://mcp.courtlistener.com",
      "auth": "oauth"
    }
  }
}
```
~/.mcporter/credentials.json includes updated tokens:
```

{
  "version": 1,
  "entries": {
    "courtlistener|746376d5ce7a26fa": {
      "serverName": "courtlistener",
      "serverUrl": "https://mcp.courtlistener.com/",
      "updatedAt": "2026-05-17T14:50:36.432Z",
      "state": "e30e0c84-21b0-4195-a565-63b582d7cad3",
      "codeVerifier": "-ZR-VMq99sU9sh.3eYEDKlCykb9kz2qI5BPOyzKt_fo",
      "clientInfo": {
        "redirect_uris": [
          "http://127.0.0.1:42197/callback"
        ],
        "token_endpoint_auth_method": "none",
        "grant_types": [
          "authorization_code",
          "refresh_token"
        ],
        "response_types": [
          "code"
        ],
        "client_name": "mcporter (courtlistener)",
        "client_id": "VALID_CLIENT_ID_REDACTED",
        "client_id_issued_at": 1779029391
      },
      "tokens": {
        "access_token": "VALID-TOKEN-REDACTED",
[...]

PR #182, opened 2026-05-20
fix(daemon): pass allowCachedAuth to runtime for OAuth token reuse

### LINKED ISSUE #181: Daemon never passes allowCachedAuth — OAuth servers fail after token expiry
## Summary

The daemon does not pass `allowCachedAuth: true` when calling `runtime.callTool()` or `runtime.listTools()` in `host.ts`, and the `KeepAliveRuntime.callTool()` wrapper in `runtime-wrapper.ts` does not forward `allowCachedAuth` to the daemon client. This means any OAuth server running through the keep-alive daemon can never use cached tokens — every call goes through `createClientContext` with `allowCachedAuth: undefined`, which skips `applyCachedAuthIfAvailable` entirely.

Combined with the `completeProactiveAuthorization` regression from v0.11.0 (issue #179), this makes daemon-managed OAuth servers completely non-functional for non-interactive use after the first token expiry.

## Impact

- All OAuth-authenticated servers configured with `lifecycle: "keep-alive"` fail once access tokens expire (~1 hour)
- The daemon cannot auto-refresh tokens because it never reads them
- Users must re-authenticate manually via browser on every call

### Bug 1: `host.ts` — Daemon never passes `allowCachedAuth`

In `processRequest()`, both `callTool` and `listTools` cases call the runtime without `allowCachedAuth`:

```typescript
// host.ts — callTool handler
const result = await runtime.callTool(params.server, params.tool, {
  args: params.args ?? {},
  timeoutMs: params.timeoutMs,
  // Missing: allowCachedAuth: true,
});

// host.ts — listTools handler  
const result = await runtime.listTools(params.server, {
  includeSchema: params.includeSchema,
  autoAuthorize: params.autoAuthorize,
  // Missing: allowCachedAuth: true,
});
```

Without `allowCachedAuth: true`, `createClientContext` calls `applyCachedAuthIfAvailable(definition, logger, undefined)`. The guard at `transport.ts:181` — `if (!allowCachedAuth && definition.auth !== 'refreshable_bearer')` — returns the definition without reading the token cache, so cached OAuth tokens are never loaded.

### Bug 2: `runtime-wrapper.ts` — `callTool` doesn't forward `allowCachedAuth`

```typescript
// runtime-wrapper.ts — callTool
async callTool(server: string, toolName: string, options?: CallOptions): Promise<unknown> {
  if (this.shouldUseDaemon(server)) {
    return this.invokeWithRestart(server, 'callTool', () =>
      this.daemon.callTool({
        server,
        tool: toolName,
        args: options?.args,
        timeoutMs: options?.timeoutMs,
        // Missing: allowCachedAuth: options?.allowCachedAuth ?? true,
      })
    );
  }
  return this.base.callTool(server, toolName, options);
}
```

The `listTools` wrapper already includes `allowCachedAuth: options?.allowCachedAuth ?? true`, but `callTool` was missed.

### Bug 3: `protocol.ts` — `CallToolParams` type missing `allowCachedAuth`

```typescript
export interface CallToolParams {
  readonly server: string;
  readonly tool: string;
  readonly args?: Record<string, unknown>;
  readonly timeoutMs?: number;
  // Missing: allowCachedAuth?: boolean;
}
```

The daemon protocol type doesn't define `allowCachedAuth`, so even if a client sends it, the daemon can't type-check it.

## Related

- **Issue #179** — `completeProactiveAuthorization` destroys working credentials by re-registering OAuth clients with mismatched redirect URIs. Even once the daemon passes `allowCachedAuth`, this second bug will trigger on the first token refresh unless both are fixed.

## Proposed Fix

```diff
// src/daemon/protocol.ts
export interface CallToolParams {
  readonly server: string;
  readonly tool: string;
  readonly args?: Record<string, unknown>;
  readonly timeoutMs?: number;
+ readonly allowCachedAuth?: boolean;
}
```

```diff
// src/daemon/host.ts — callTool case in processRequest()
const result = await runtime.callTool(params.server, params.tool, {
  args: params.args ?? {},
  timeoutMs: params.timeoutMs,
+ allowCachedAuth: params.allowCachedAuth ?? true,
});
```

```diff
// src/daemon/host.ts — listTools case in processRequest()
const result = await runtime.listTools(params.server, {
  includeSchema: params.includeSchema,
  autoAuthorize: params.autoAuthorize,
+ allowCachedAuth: true,
});
```

```diff
// src/daemon/runtime-wrapper.ts — callTool
this.daemon.callTool({
  server,
  tool: toolName,
  args: options?.args,
  timeoutMs: options?.timeoutMs,
+ allowCachedAuth: options?.allowCachedAuth ?? true,
})
```

## Environment

- mcporter 0.11.1
- macOS 26.5 (Apple Silicon)
- 15 keep-alive OAuth servers (Cloudflare, Composio, Snoooz, MoreGoodReviews, Puzzle, Voicenotes)
- All fail after ~1 hour when access tokens expire
- **Workaround**: manually patched `dist/daemon/host.js` and `dist/daemon/runtime-wrapper.js` to hardcode `allowCachedAuth: true` (wiped on `npm update`)

bradhallett:
## Problem

The daemon never passes `allowCachedAuth` when calling `runtime.callTool()` or `runtime.listTools()` in `host.ts`, and the `KeepAliveRuntime.callTool()` wrapper does not forward it to the daemon client. This means any OAuth server running through the keep-alive daemon can never use cached tokens.

Without `allowCachedAuth`, `createClientContext` calls `applyCachedAuthIfAvailable(definition, logger, undefined)`. The guard at `transport.ts:181` — `if (!allowCachedAuth && definition.auth !== 'refreshable_bearer')` — returns early without reading the token cache, so cached OAuth tokens are never loaded.

Combined with the `completeProactiveAuthorization` regression (#179), this makes daemon-managed OAuth servers completely non-functional for non-interactive use after the first token expiry (~1 hour).

## Changes

4 lines across 3 files:

```diff
src/daemon/protocol.ts        | 1 +
src/daemon/host.ts            | 2 ++
src/daemon/runtime-wrapper.ts | 1 +
```

- **protocol.ts**: Add `allowCachedAuth?: boolean` to `CallToolParams`
- **host.ts**: Pass `allowCachedAuth` in both `callTool` and `listTools` handlers
- **runtime-wrapper.ts**: Forward `allowCachedAuth` in `callTool` daemon delegation

## Testing

Workaround verified in production — manually patched `dist/daemon/host.js` and `dist/daemon/runtime-wrapper.js` with the same changes. All 15 keep-alive OAuth servers (Cloudflare, Composio, Snoooz, MoreGoodReviews, Puzzle, Voicenotes) have been stable for 48+ hours with automatic token refresh.

Fixes #181
Related #179

### DIFF
--- a/src/cli/tool-cache.ts
+++ b/src/cli/tool-cache.ts
@@ -12,7 +12,7 @@ const runtimeCache = new WeakMap<Runtime, Map<string, Promise<ToolMetadata[]>>>(
 function cacheKey(serverName: string, options: LoadToolMetadataOptions): string {
   const includeSchema = options.includeSchema !== false;
   const autoAuthorize = options.autoAuthorize !== false;
-  const allowCachedAuth = options.allowCachedAuth === true;
+  const allowCachedAuth = options.allowCachedAuth !== false;
   return `${serverName}::schema:${includeSchema ? '1' : '0'}::auth:${autoAuthorize ? '1' : '0'}::cached-auth:${allowCachedAuth ? '1' : '0'}`;
 }
 
@@ -33,10 +33,11 @@ export async function loadToolMetadata(
   }
   const includeSchema = options.includeSchema !== false;
   const autoAuthorize = options.autoAuthorize !== false;
-  const listOptions: ListToolsOptions =
-    options.allowCachedAuth === undefined
-      ? { includeSchema, autoAuthorize }
-      : { includeSchema, autoAuthorize, allowCachedAuth: options.allowCachedAuth };
+  const listOptions: ListToolsOptions = {
+    includeSchema,
+    autoAuthorize,
+    allowCachedAuth: options.allowCachedAuth ?? true,
+  };
   const promise = runtime
     .listTools(serverName, listOptions)
     .then((tools) => tools.map((tool) => buildToolMetadata(tool)))
--- a/src/daemon/host.ts
+++ b/src/daemon/host.ts
@@ -302,6 +302,7 @@ async function processRequest(
           const result = await runtime.callTool(params.server, params.tool, {
             args: params.args ?? {},
             timeoutMs: params.timeoutMs,
+            allowCachedAuth: params.allowCachedAuth ?? true,
           });
           markActivity(params.server, activity);
           if (loggable) {
@@ -327,6 +328,7 @@ async function processRequest(
           const result = await runtime.listTools(params.server, {
             includeSchema: params.includeSchema,
             autoAuthorize: params.autoAuthorize,
+            allowCachedAuth: true,
           });
           markActivity(params.server, activity);
           if (loggable) {
--- a/src/daemon/host.ts
+++ b/src/daemon/host.ts
@@ -302,7 +302,6 @@ async function processRequest(
           const result = await runtime.callTool(params.server, params.tool, {
             args: params.args ?? {},
             timeoutMs: params.timeoutMs,
-            allowCachedAuth: params.allowCachedAuth ?? true,
           });
           markActivity(params.server, activity);
           if (loggable) {
@@ -328,7 +327,7 @@ async function processRequest(
           const result = await runtime.listTools(params.server, {
             includeSchema: params.includeSchema,
             autoAuthorize: params.autoAuthorize,
-            allowCachedAuth: true,
+            allowCachedAuth: params.allowCachedAuth ?? true,
           });
           markActivity(params.server, activity);
           if (loggable) {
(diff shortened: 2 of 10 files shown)


steipete:
Maintainer fixup pushed on top of this PR.

What changed after review:
- `callTool` now uses cached OAuth credentials by default in the runtime, including daemon-backed calls.
- `listTools` keeps an explicit opt-out for callers that need uncached auth behavior.
- runtime connection caching now reconnects when cached-auth mode changes, so an earlier uncached connection cannot mask a later cached-auth tool call.
- added regression coverage for daemon host/wrapper paths and runtime cached-token behavior.

Local proof:
- `git diff --check`
- `./runner pnpm check`
- `./runner pnpm test`

GitHub Socket checks are green on the pushed PR head.

PR #198, opened 2026-06-06
fix(runtime): preserve disableOAuth across headless paths

### LINKED ISSUE #197: Add `disableOAuth` connect option — `maxOAuthAttempts: 0` defeats the connection cache
# Add `disableOAuth` connect option — `maxOAuthAttempts: 0` defeats the connection cache

## Summary

`Runtime.connect()` currently uses `options.maxOAuthAttempts === undefined` as part of its cache-key check. Callers who want to suppress the interactive OAuth flow (typically long-running headless daemons that have no browser and rely on cached tokens only) must pass `maxOAuthAttempts: 0`, which has the side effect of permanently disabling connection caching for that call site. Every subsequent `callTool` / `listTools` spawns a fresh transport that `runtime.close()` cannot reap.

There's no first-class way to say "I never want OAuth to start, but please keep pooling connections."

## Reproduction

Reproduced against `mcporter@0.11.0` locally (please confirm against current `main` / v0.11.3). The relevant section of `dist/runtime.js` is unchanged in 0.11.3 as far as I can tell from the published source.

```js
// dist/runtime.js, around the connect() method:
async connect(server, options = {}) {
  const normalized = server.trim();
  const useCache = options.skipCache !== true && options.maxOAuthAttempts === undefined;
  if (useCache) {
    const existing = this.clients.get(normalized);
    if (existing) return existing;
  }
  // ... createClientContext(...) ...
  if (useCache) {
    this.clients.set(normalized, connection);
    // ...
  }
  return connection;
}
```

Minimal repro (TypeScript pseudo):

```ts
import { createRuntime } from "mcporter";

const runtime = await createRuntime(/* loaded definitions */);

// Headless daemon use case: never start OAuth. The documented way to do
// this is maxOAuthAttempts: 0.
const wrappedConnect = (server, options = {}) =>
  runtime.connect(server, { ...options, maxOAuthAttempts: 0, allowCachedAuth: true });

// Override callTool/listTools to use wrappedConnect, then poll.
for (let i = 0; i < 100; i++) {
  await runtime.callTool("some-server", "some-tool", { args: {} });
}

await runtime.close();
// Expected: one cached client created at i=0, reused for i=1..99, closed by runtime.close().
// Actual:   100 transports spawned (one per callTool), none cached, none closed by runtime.close().
```

The `runtime.close()` loop iterates `this.clients.entries()`; with `useCache === false` on every call, nothing ever lands in that Map.

## Impact

- **fd / process leak under sustained load.** A daemon polling N accounts every M seconds spawns N×(3600/M)×24 transports per day. For N=20 and M=30s that's about 5,800/day, none reaped until process exit.
- **Latency regression.** Each `callTool` pays the full transport-establishment cost (STDIO spawn + handshake, or HTTP+OAuth-context bootstrap) on every invocation instead of amortizing across the cache.
- **Resilience cost.** `resetConnectionOnError` is keyed on the same `this.clients` map (`if (!this.clients.has(normalized)) return;`), so error-path cleanup is also a no-op for these connections.

## Why the conflation is hard to work around at the caller layer

The caller wants three things simultaneously:

1. `shouldEstablishOAuth` evaluates false → no browser, no interactive flow.
2. `applyCachedAuthIfAvailable` runs → cached tokens get a fair shot.
3. Connections are pooled across calls.

Today only (1) and (2) are reachable as a combination — selecting them via `maxOAuthAttempts: 0 + allowCachedAuth: true` toggles (3) off as a side effect.

Caller-side workarounds:

- **Manage our own cache wrapping `connect`.** Feasible but requires understanding mcporter's `{client}` shape, what `runtime.close()` is doing internally beyond client-level close, and tracking changes across mcporter versions. We've considered this; it's brittle.
- **Set `maxOAuthAttempts: 1` instead of `0`.** Doesn't help — `useCache` is gated on `=== undefined`, not on `=== 0`. Any non-undefined value still bypasses the cache, and `1` re-introduces the OAuth flow we needed to suppress.
- **Patch `connect` via prototype monkey-patch on the runtime.** That's the daemon's actual current shape, but it's exactly why this issue exists — the patch can't restore caching without re-implementing the cache itself.

## Proposed API

Add a `disableOAuth: boolean` option (or another name the maintainers prefer) that says "never start an OAuth flow for this server, regardless of `def.auth` setting." It should NOT affect cache behavior.

```ts
interface ConnectOptions {
  skipCache?: boolean;
  maxOAuthAttempts?: number;
  allowCachedAuth?: boolean;
  oauthSessionOptions?: OAuthSessionOptions;
  // NEW:
  disableOAuth?: boolean;
}
```

Behavior changes inside `connect`:

```js
async connect(server, options = {}) {
  const normalized = server.trim();
  // disableOAuth is cache-friendly: connections established with it are
  // identical to default-OAuth-suppressed connections, so they share a
  // cached entry.
  const useCache =
    options.skipCache !== true &&
    options.maxOAuthAttempts === undefined &&
    // disableOAuth is opt-in and never disables caching:
    true;
  // ... rest unchanged ...
}
```

And `shouldEstablishOAuth` in `createClientContext` (or wherever it lives) gains a clause:

```js
const shouldEstablishOAuth =
  activeDefinition.auth === "oauth" &&
  options.maxOAuthAttempts !== 0 &&
  options.disableOAuth !== true;
```

(I haven't traced exactly where `shouldEstablishOAuth` is computed — if it's in `createClientContext` vs `transport.js` the diff lands in a slightly different place.)

## Backward compatibility

- Existing callers passing `maxOAuthAttempts: 0` keep their current behavior (no cache, no OAuth). The new option is purely additive.
- Existing callers passing `skipCache: true` keep their current behavior (no cache, OAuth flow allowed). Unchanged.
- New behavior is opt-in via the new flag.

No version bump beyond a minor is needed.

## Alternative the maintainers might prefer

If `disableOAuth` is too narrow a name, naming alternatives that have come up internally:

- `oauthMode: "cached-only" | "interactive" | "off"` — more explicit, lets future modes slot in.
- `headless: true` — matches the daemon use case, but might be confusing for non-daemon callers who happen to want OAuth suppression.
- Document that callers wanting OAuth-off-with-cache should pass `{ maxOAuthAttempts: 0, _allowCache: true }` and update the `useCache` check to honor a new `_allowCache` flag.

I'm happy with whichever shape the maintainers prefer. Glad to put up a PR once the API direction is settled.

## Use case context

This came up implementing a long-running daemon process that polls a hosted MCP server every 30 seconds across multiple OAuth-credentialed accounts. The daemon has no GUI, so any OAuth flow that starts is a hard failure mode (`open ./oauth.html` is a no-op in launchd-managed processes, leaving the daemon hung waiting for a redirect that never arrives). The cached-tokens-only path via `maxOAuthAttempts: 0` works correctly for OAuth suppression, but the resulting transport leak makes the daemon unsuitable for production over multi-day soak.

Happy to share more concrete numbers from soak testing if useful.

## Related

- `dist/runtime.js` — `connect()` method, `useCache` calculation
- `dist/runtime.js` — `close()` method, iteration over `this.clients`
- `dist/runtime.js` — `resetConnectionOnError()`, also keyed on `this.clients`
- mcporter's OAuth-flow path (`transport.js` / `oauth.js`?)

feniix:
## Summary

Headless MCPorter callers can now suppress interactive OAuth without losing the connection pooling they need for daemon and repeated-tool workflows. `disableOAuth` now stays intact across direct runtime helpers, keep-alive daemon paths, and CLI commands instead of only working at the low-level `connect()` entry point.

This closes openclaw/mcporter#197.

## What changed

- Preserves cached no-OAuth connections through `callTool`, `listTools`, `listResources`, and `readResource`, including the case where a caller pre-connects with `disableOAuth: true` and then uses higher-level helpers.
- Forwards the option through keep-alive runtime wrappers, daemon protocol/host handling, and daemon fast-path calls so daemon-managed servers cannot accidentally re-enable OAuth.
- Exposes `--disable-oauth` / `--no-oauth` for `mcporter call`, `mcporter list`, and `mcporter resource`.
- Normalizes `disableOAuth: false` to the same cache identity as omitting the option.
- Adds regression coverage for OAuth server suppression, 401 promotion suppression, helper reuse, daemon forwarding, CLI parsing, and tool metadata cache keys.

## Demo

A local terminal evidence capture was recorded at:

`/var/folders/l7/gpl83h991dbfbmbr1v8860f00000gn/T/mcporter-disable-oauth-demo-XXXXXX.KvYUH5Y2tf/disable-oauth-cli-evidence.svg`

It shows the actual built CLI help for `call`, `list`, and `resource` exposing `--disable-oauth`.

## Test plan

- `./runner pnpm run check`
- `./runner pnpm test`
- `./runner pnpm exec tsx scripts/docs-list.ts`
- `./git diff --check`

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Pi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### OAuth
 
+- Add cache-friendly `disableOAuth` support across headless runtime, CLI, daemon, proxy, and `callOnce` paths so callers can suppress interactive OAuth without losing connection reuse. (Issues #197, #199, #201, thanks @feniix)
 - Recover cleanly from renamed OAuth server entries, invalid refresh tokens, and stale dynamic client registrations without reusing unrelated same-URL credentials.
 
 ### CLI
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -38,6 +38,8 @@ A quick reference for the primary `mcporter` subcommands. Each command inherits
   - `--exit-code` – exit 1 when any checked server is unhealthy.
   - `--quiet` – suppress output and exit 1 when any checked server is unhealthy.
   - `--timeout <ms>` – per-server timeout when enumerating all servers.
+  - `--disable-oauth` – never start an interactive OAuth flow; use cached
+    tokens only while keeping eligible connections pooled.
 
 ## `mcporter call <server.tool>`
 
@@ -52,6 +54,8 @@ A quick reference for the primary `mcporter` subcommands. Each command inherits
   - `--raw-strings` – disable numeric coercion for flag-style and positional values.
   - `--no-coerce` – disable all flag-style/positional value coercion.
   - `--tail-log` – stream tail output when the tool returns log handles.
+  - `--disable-oauth` – never start an interactive OAuth flow; use cached
+    tokens only while keeping eligible connections pooled.
 
 ## `mcporter resource <server> [uri]`
 
@@ -63,6 +67,8 @@ A quick reference for the primary `mcporter` subcommands. Each command inherits
   - `--output auto|text|markdown|json|raw` – choose how to render the response.
   - `--json` – shortcut for `--output json`.
   - `--raw` – shortcut for `--output raw`.
+  - `--disable-oauth` – never start an interactive OAuth flow; use cached
+    tokens only while keeping eligible connections pooled.
 
 ## `mcporter serve [--servers a,b,c] [--stdio | --http <port>]`
 
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -38,7 +38,7 @@ A quick reference for the primary `mcporter` subcommands. Each command inherits
   - `--exit-code` – exit 1 when any checked server is unhealthy.
   - `--quiet` – suppress output and exit 1 when any checked server is unhealthy.
   - `--timeout <ms>` – per-server timeout when enumerating all servers.
-  - `--disable-oauth` – never start an interactive OAuth flow; use cached
+  - `--no-oauth` – never start an interactive OAuth flow; use cached
     tokens only while keeping eligible connections pooled.
 
 ## `mcporter call <server.tool>`
@@ -54,7 +54,7 @@ A quick reference for the primary `mcporter` subcommands. Each command inherits
   - `--raw-strings` – disable numeric coercion for flag-style and positional values.
   - `--no-coerce` – disable all flag-style/positional value coercion.
   - `--tail-log` – stream tail output when the tool returns log handles.
-  - `--disable-oauth` – never start an interactive OAuth flow; use cached
+  - `--no-oauth` – never start an interactive OAuth flow; use cached
     tokens only while keeping eligible connections pooled.
 
 ## `mcporter resource <server> [uri]`
@@ -67,7 +67,7 @@ A quick reference for the primary `mcporter` subcommands. Each command inherits
   - `--output auto|text|markdown|json|raw` – choose how to render the response.
   - `--json` – shortcut for `--output json`.
   - `--raw` – shortcut for `--output raw`.
-  - `--disable-oauth` – never start an interactive OAuth flow; use cached
+  - `--no-oauth` – never start an interactive OAuth flow; use cached
     tokens only while keeping eligible connections pooled.
 
 ## `mcporter serve [--servers a,b,c] [--stdio | --http <port>]`
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -35,7 +35,7 @@ Use `createServerProxy(runtime, name)` inside scripts when you want ergonomic ca
 2. Automatically merges default values.
 3. Returns a `CallResult` helper so you can render `.text()`, `.markdown()`, or `.json()` without manual parsing.
 
-When you need raw access (custom transports, streaming), use the bare `Client` from `@modelcontextprotocol/sdk` or inspect `runtime.connect(name)` for lower-level control. Headless callers that must rely on cached tokens without launching OAuth can pass `disableOAuth: true` to `connect`, `callTool`, `listTools`, and resource helpers; this suppresses interactive OAuth while keeping eligible connections pooled.
+When you need raw access (custom transports, streaming), use the bare `Client` from `@modelcontextprotocol/sdk` or inspect `runtime.connect(name)` for lower-level control. Headless callers that must rely on cached tokens without launching OAuth can pass `disableOAuth: true` to `connect`, `callTool`, `listTools`, resource helpers, and `callOnce`; this suppresses interactive OAuth while keeping eligible connections pooled.
 
 ## Debug + Support Docs
 
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -35,7 +35,7 @@ Use `createServerProxy(runtime, name)` inside scripts when you want ergonomic ca
 2. Automatically merges default values.
 3. Returns a `CallResult` helper so you can render `.text()`, `.markdown()`, or `.json()` without manual parsing.
 
-When you need raw access (custom transports, streaming), use the bare `Client` from `@modelcontextprotocol/sdk` or inspect `runtime.connect(name)` for lower-level control.
+When you need raw access (custom transports, streaming), use the bare `Client` from `@modelcontextprotocol/sdk` or inspect `runtime.connect(name)` for lower-level control. Headless callers that must rely on cached tokens without launching OAuth can pass `disableOAuth: true` to `connect`, `callTool`, `listTools`, and resource helpers; this suppresses interactive OAuth while keeping eligible connections pooled.
 
 ## Debug + Support Docs
 
--- a/examples/headless-pooling-demo.ts
+++ b/examples/headless-pooling-demo.ts
@@ -1,0 +1,171 @@
+#!/usr/bin/env tsx
+
+/**
+ * Demonstration: `disableOAuth: true` provides cache-friendly OAuth
+ * suppression for headless callers.
+ *
+ * Spins up a local mock MCP server (no real auth), then exercises three
+ * patterns side-by-side and counts the distinct ClientContext objects
+ * the runtime hands out:
+ *
+ *   1. Legacy `maxOAuthAttempts: 0`         — uncached (existing contract).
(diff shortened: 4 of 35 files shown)


feniix:
# Real-behavior proof — PR #198

Two artifacts: vitest output showing the regression tests pass, and a
runnable demo (`examples/headless-pooling-demo.ts`) that demonstrates the
three patterns against a local mock MCP server.

Reproducible from this branch:

```
$ pnpm test -- tests/runtime-integration.test.ts --reporter=verbose
$ pnpm tsx examples/headless-pooling-demo.ts
```

## 1. Regression tests (vitest)

The integration test suite carries 8 cases that lock in the new
`disableOAuth` contract. All pass.

```
 RUN  v4.1.5 /Users/feniix/src/spantree/mcporter

 ✓ tests/runtime-integration.test.ts > runtime integration > lists tools and calls a tool over HTTP 27ms
 ✓ tests/runtime-integration.test.ts > runtime integration > lists and reads resources over HTTP 4ms
 ✓ tests/runtime-integration.test.ts > runtime integration > reuses cached connection when disableOAuth: true is passed 6ms
 ✓ tests/runtime-integration.test.ts > runtime integration > treats disableOAuth: false like omitted for cache identity 2ms
 ✓ tests/runtime-integration.test.ts > runtime integration > maxOAuthAttempts: 0 still bypasses the cache (existing contract preserved) 2ms
 ✓ tests/runtime-integration.test.ts > runtime integration > evicts and re-establishes the cached client when disableOAuth flag changes 3ms
 ✓ tests/runtime-integration.test.ts > runtime integration > preserves the cached client across connect(disableOAuth:true) → callTool() (no implicit eviction) 3ms
 ✓ tests/runtime-integration.test.ts > runtime integration > preserves the cached client across connect(disableOAuth:true) → listTools() (no implicit eviction) 4ms

 Test Files  1 passed (1)
      Tests  8 passed (8)
```

The last two cases are the targeted regression for review comment
[r3366238654](https://github.com/openclaw/mcporter/pull/198#discussion_r3366238654)
— they prove that `connect(disableOAuth: true)` followed by `callTool()`
or `listTools()` does NOT evict the pre-connected slot.

Full suite (all 124 test files): **723 pass / 3 skipped / 0 fail**.
`pnpm lint` + `pnpm typecheck` clean.

## 2. Demonstrating the three patterns end-to-end

`examples/headless-pooling-demo.ts` starts a local express + MCP mock
server with a public `add` tool, then drives the runtime through three
patterns side-by-side, counting distinct ClientContext objects via
identity.

```
$ pnpm tsx examples/headless-pooling-demo.ts
[demo] Mock MCP server listening at http://127.0.0.1:55082/mcp

[demo] Pattern A — legacy maxOAuthAttempts: 0
[demo]   5 connect() calls → 5 distinct ClientContexts
[demo]   Expected: 5 (legacy contract: cache disabled when maxOAuthAttempts is set)
[demo]   Result:   OK

[demo] Pattern B — disableOAuth: true on every connect
[demo]   5 connect() calls → 1 distinct ClientContexts
[demo]   Expected: 1 (cache reuse under cache-friendly suppression)
[demo]   Result:   PASS

[demo] Pattern C — pre-connect(disableOAuth:true) + 5 callTool()
[demo]   Sum of 5 add() results: 25
[demo]   Post-callTool connect() === pre-connect ClientContext: true
[demo]   Expected: true (no implicit eviction from callTool internals)
[demo]   Result:   PASS
```

What each pattern proves:

* **Pattern A** (existing contract preserved): callers that opt out via
  the legacy `maxOAuthAttempts: 0` knob still bypass the cache. No
  behavior change for existing code.
* **Pattern B** (the new contract works): repeated `connect()` calls
  with `disableOAuth: true` return the same cached ClientContext, so
  the daemon use case — `wrapRuntimeForDaemon` forcing this option on
  every internal `connect` — actually pools transports instead of
  spawning one per RPC.
* **Pattern C** (the review-comment regression): the documented setup
  `await runtime.connect(server, { disableOAuth: true })` followed by
  `callTool()` keeps the pre-connected slot. The 5 `add()` invocations
  succeed (sum = 0+1 + 1+2 + 2+3 + 3+4 + 4+5 = 1+3+5+7+9 = 25) and the
  post-callTool `connect()` returns `=== ` the original ClientContext.

OAuth-suppression semantics under `disableOAuth: true` (the
`shouldEstablishOAuth` / `maybePromoteHttpDefinition` short-circuits)
are exercised by the existing unit tests in
`tests/runtime-transport.test.ts`. The demo's mock server has no
`auth: 'oauth'` definition so it doesn't exercise the OAuth code path —
proving the cache fix end-to-end on a real transport was the goal here.

feniix:
@clawsweeper re-review

feniix:
Addressed the proxy schema-discovery gap from the last review.

`dfd1b6d` (`fix(server-proxy): thread disableOAuth through schema-discovery listTools`) threads `disableOAuth` from the caller's options into the proxy's pre-call `runtime.listTools(server, { includeSchema: true })` so the schema-fetch and the eventual `callTool` share the same no-OAuth posture. End-to-end no-browser guarantee now holds across the proxy interface.

Regression test in `tests/server-proxy.test.ts` asserts both:

- `runtime.listTools('mock', { includeSchema: true, disableOAuth: true })`
- `runtime.callTool('mock', 'some-tool', { args: { foo: 'bar' }, disableOAuth: true })`

`pnpm test`: 724 pass / 3 skipped / 0 fail. `pnpm check` (format + lint + typecheck) clean.

@clawsweeper re-review

steipete:
Maintainer verification on `f837500`:

- Focused proxy/runtime suite: 9 files, 81 tests passed.
- `pnpm check`: format, lint, and typecheck passed.
- `pnpm test`: 125 files passed, 1 skipped; 773 tests passed, 3 skipped.
- `pnpm tsx examples/headless-pooling-demo.ts`: legacy uncached, cache-friendly suppression, and pre-connect reuse patterns all passed.
- Local patch autoreview: clean.
- Full branch autoreview against `origin/main`: clean.
- GitHub CI: Ubuntu, macOS, Windows, and both Socket checks passed.

The final proxy handling preserves schema-owned option names, uses uncached no-authorize discovery for ambiguous argument shapes, and isolates concurrent schema discovery by OAuth posture.

PR #208, opened 2026-06-16
fix(oauth): degrade to re-auth on corrupt credential cache instead of crashing

### LINKED ISSUE #207: [Bug]: corrupt OAuth credential cache file crashes connection instead of degrading to re-auth (DirectoryPersistence)
### Bug type

Error handling / resilience — a present-but-corrupt credential cache file crashes an MCP connection instead of degrading to re-auth.

### Summary

`DirectoryPersistence` (used whenever a server sets `tokenCacheDir`, and during the best-effort legacy-cache migration) reads OAuth artifacts through `readJsonFile`, whose `catch` only handles `ENOENT`:

```ts
// src/fs-json.ts:14
export async function readJsonFile<T = unknown>(filePath: string): Promise<T | undefined> {
  try {
    const content = await fs.readFile(filePath, 'utf8');
    return JSON.parse(content) as T;
  } catch (error) {
    if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
      return undefined;
    }
    throw error; // <-- a SyntaxError from a corrupt file is re-thrown
  }
}
```

So a **present-but-corrupt** token/client/state file (truncated mid-write by a killed process, disk-full on an older build, hand-edited, etc.) makes `JSON.parse` throw a `SyntaxError`, which propagates straight out of the provider's `tokens()` / `clientInformation()` and surfaces as a hard connection error — rather than being treated as "no usable record → re-authenticate".

### Why this looks like a defect (in-repo evidence, not an assumption)

The same codebase already treats a corrupt persisted file as a recoverable miss **everywhere else it matters** — `DirectoryPersistence` is the lone holdout:

- `VaultPersistence` (the default store) → `loadVaultEntry` → `readVaultState` (`src/oauth-vault.ts:39`) **explicitly catches `SyntaxError`** and returns an empty, `needsRepair: true` vault.
- `daemon/host.ts:258` `metadataMatches` wraps the read in `catch { return false }`.
- `server-proxy.ts:336` `loadSchemasFromDisk` wraps `readSchemaCache` in `catch { /* ignore cache read failures */ }`.
- `DirectoryPersistence.readTokens/readClientInfo/readState` (`src/oauth-persistence.ts:154,164,189`) → return `readJsonFile(...)` directly, with **no `SyntaxError` handling**.

The provider contract is `Promise<OAuthTokens | undefined>` — `undefined` already means "no creds, go authenticate". A corrupt file is functionally "no usable creds", so the default (vault) path repairs and re-auths, while a user who configured `tokenCacheDir` gets a crash instead. That asymmetry is the bug.

### Affected paths

1. **Runtime token read.** `OAuthClientProvider.tokens()` (`src/oauth.ts:264`) → `CompositePersistence.readTokens()` (`src/oauth-persistence.ts:277`, no try/catch around `store.readTokens()`) → `DirectoryPersistence.readTokens()` → `readJsonFile` throws → connection fails instead of re-authenticating.
2. **Legacy migration.** `buildOAuthPersistence` (`src/oauth-persistence.ts:350-353`) calls `legacy.readTokens()/readClientInfo()/readState()` on the legacy `~/.mcporter/<name>` dir. A single corrupt legacy file throws and aborts the *best-effort* migration — and therefore server setup — even though migration is meant to be tolerant.

### Steps to reproduce (analysis-level — not run in a sandbox; see "What was not verified")

1. Configure a server with a `tokenCacheDir`.
2. Authenticate once so `<tokenCacheDir>/tokens.json` (or `client.json`/`state.json`) is written.
3. Corrupt that file (e.g. truncate it, or write `{`).
4. Trigger a connection that calls the OAuth provider's `tokens()`.

**Expected:** corrupt cache treated as "no creds" → silent re-auth (matching `VaultPersistence`).
**Actual (from source trace):** `SyntaxError: Unexpected end of JSON input` propagates out of `tokens()` and the connection fails.

### Proposed fix (options)

- **A (preferred, narrow & consistent):** make `DirectoryPersistence`'s three readers degrade on parse failure exactly like `VaultPersistence`/`daemon` already do — catch `SyntaxError` (keep re-throwing genuine I/O faults) and return `undefined`. Sweep all three sibling readers in one change, not just `readTokens`.
- **B:** add an opt-in `tolerateCorrupt` flag to `readJsonFile` and pass it from the cache/credential readers (leaves the strict default for callers that want it). *Note:* do **not** make `readJsonFile` swallow parse errors unconditionally — `oauth-vault.ts` relies on distinguishing a `SyntaxError` (repair) from a missing file, so a blanket change there would regress the vault's repair logic.

### What was not verified

This is a **static source analysis**, not an observed crash — I have not built mcporter or executed the OAuth flow against a corrupt file. The `SyntaxError`-propagation path is traced through the code above and corroborated by the in-repo sibling handlers that deliberately recover from the same condition. Happy to be told if the throw is intended (e.g. you'd rather surface credential-file tampering loudly) — in which case the inconsistency with `VaultPersistence` is itself worth a note.

---
*Filed with AI assistance (static analysis). Cross-repo note: the analogous corrupt-session-read recovery was just accepted in `openclaw/acpx` (#379).*

KrasimirKralev:
## Problem

When a cached OAuth credential file under an explicit `tokenCacheDir` (or a legacy per-server cache during migration) is present but **corrupt** — truncated or malformed JSON, e.g. from an interrupted write — `DirectoryPersistence`'s readers let the `JSON.parse` `SyntaxError` propagate. Through `CompositePersistence` that crashes the MCP connection instead of treating the file as "no usable credentials" and degrading to re-auth.

## Root cause

`DirectoryPersistence.readTokens()` and `readClientInfo()` call `readJsonFile`, which maps only `ENOENT` to `undefined` and re-throws everything else (including `SyntaxError`).

This is asymmetric with the rest of the codebase, which already treats a corrupt credential file as recoverable:

- `VaultPersistence` / `readVaultState` (`oauth-vault.ts`) catches `SyntaxError` and rebuilds (`needsRepair`)
- `daemon/host.ts` and `server-proxy.ts` wrap their reads in catch-all handlers

These `DirectoryPersistence` readers were the lone exceptions.

## Fix

Route the **credential** readers (`readTokens`, `readClientInfo`) through a small private helper (`readJsonOrUndefined`) that maps **only** `SyntaxError` → `undefined`; genuine I/O faults still propagate. Mirrors the existing `VaultPersistence` pattern (`if (!(error instanceof SyntaxError)) throw error`).

**OAuth `state` is deliberately excluded.** A corrupt `state.txt` must fail the flow *closed*: `oauth.ts` only enforces the callback state check when `expectedState` is truthy, so degrading state to `undefined` would skip the CSRF check (thanks @chatgpt-codex-connector for catching this on the first pass). `readState()` therefore keeps throwing on corrupt input. `readJsonFile` itself is also left untouched, since the vault relies on it throwing `SyntaxError` to distinguish corrupt-from-missing for its repair path.

## Verification

Run locally (Node 22, vitest):

- New regression test in `tests/oauth-persistence.test.ts` writes corrupt `tokens.json` / `client.json` / `state.txt` and asserts `readTokens()` / `readClientInfo()` resolve to `undefined` **while `readState()` rejects with `SyntaxError`** (CSRF fail-closed).
  - The credential-degrade assertions fail on `main` (`SyntaxError: Unterminated string in JSON`); they pass with the fix.
- Full `oauth-persistence` suite: **35/35 passing**.
- `tsgo --noEmit`, `oxfmt --check`, `oxlint --type-aware`: clean.

Fixes #207.

### DIFF
--- a/src/oauth-persistence.ts
+++ b/src/oauth-persistence.ts
@@ -152,7 +152,7 @@ class DirectoryPersistence implements OAuthPersistence {
   }
 
   async readTokens(): Promise<OAuthTokens | undefined> {
-    return readJsonFile<OAuthTokens>(this.tokenPath);
+    return this.readJsonOrUndefined<OAuthTokens>(this.tokenPath);
   }
 
   async saveTokens(tokens: OAuthTokens): Promise<void> {
@@ -162,7 +162,7 @@ class DirectoryPersistence implements OAuthPersistence {
   }
 
   async readClientInfo(): Promise<OAuthClientInformationMixed | undefined> {
-    return readJsonFile<OAuthClientInformationMixed>(this.clientInfoPath);
+    return this.readJsonOrUndefined<OAuthClientInformationMixed>(this.clientInfoPath);
   }
 
   async saveClientInfo(info: OAuthClientInformationMixed): Promise<void> {
@@ -187,7 +187,23 @@ class DirectoryPersistence implements OAuthPersistence {
   }
 
   async readState(): Promise<string | undefined> {
-    return readJsonFile<string>(this.statePath);
+    return this.readJsonOrUndefined<string>(this.statePath);
+  }
+
+  // A present-but-corrupt cache file (truncated or malformed JSON) means "no
+  // usable credentials": degrade to re-auth instead of crashing the connection,
+  // mirroring VaultPersistence and the daemon/server-proxy readers. Genuine I/O
+  // faults still propagate (readJsonFile re-throws everything except ENOENT).
+  private async readJsonOrUndefined<T>(filePath: string): Promise<T | undefined> {
+    try {
+      return await readJsonFile<T>(filePath);
+    } catch (error) {
+      if (!(error instanceof SyntaxError)) {
+        throw error;
+      }
+      this.logger?.debug?.(`Ignoring corrupt OAuth cache file ${filePath}: ${error.message}`);
+      return undefined;
+    }
   }
 
   async saveState(value: string): Promise<void> {
--- a/src/oauth-persistence.ts
+++ b/src/oauth-persistence.ts
@@ -187,13 +187,19 @@ class DirectoryPersistence implements OAuthPersistence {
   }
 
   async readState(): Promise<string | undefined> {
-    return this.readJsonOrUndefined<string>(this.statePath);
+    // Deliberately NOT corrupt-tolerant: a corrupt OAuth state must fail the
+    // flow closed. Returning undefined here would skip the CSRF state check on
+    // the authorization callback (see oauth.ts), so only the credential caches
+    // (tokens/client) degrade to re-auth.
+    return readJsonFile<string>(this.statePath);
   }
 
-  // A present-but-corrupt cache file (truncated or malformed JSON) means "no
-  // usable credentials": degrade to re-auth instead of crashing the connection,
+  // A present-but-corrupt credential cache (tokens/client) means "no usable
+  // credentials": degrade to re-auth instead of crashing the connection,
   // mirroring VaultPersistence and the daemon/server-proxy readers. Genuine I/O
   // faults still propagate (readJsonFile re-throws everything except ENOENT).
+  // OAuth state is intentionally excluded (see readState) so its CSRF check
+  // still fails closed on a corrupt state file.
   private async readJsonOrUndefined<T>(filePath: string): Promise<T | undefined> {
     try {
       return await readJsonFile<T>(filePath);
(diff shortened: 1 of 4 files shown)


KrasimirKralev:
Good catch, thanks — addressed in 94e66b1.

You're right that `readState()` must not degrade. `oauth.ts` only enforces the state check when `expectedState` is truthy, so a corrupt `state.txt` returning `undefined` would skip the CSRF check on the callback. I've narrowed the corrupt-tolerance to the credential caches only (`readTokens` / `readClientInfo`, which degrade to re-auth); `readState()` keeps throwing so the flow fails closed. The regression test now asserts `readState()` rejects with `SyntaxError` while tokens/client resolve to `undefined`.

vincentkoc:
Maintainer verification for bea0ed6:

- Corrupt token and client caches degrade to re-auth; corrupt OAuth state remains fail-closed.
- Stale-client I/O failures still close the callback server.
- Focused OAuth coverage, format/lint/typecheck, and the full suite passed: 779 passed, 3 skipped.
- Fresh local Codex review found no actionable findings.

@codex review

PR #241, opened 2026-07-27
fix(auth): keep the OAuth listTools request alive across every phase

umutkeltek:
Resubmission of #237 with the correctness gap fixed and a clean history, now updated for the Codex and ClawSweeper review round.

## The gap in #237

`resolveOperationSocketBudget` sized the `listTools` socket at `2 * timeoutMs + 5s`: one phase for OAuth, one for a single `tools/list`. But `McpRuntime.listTools()` walks `nextCursor` and gives **every** page its own `timeoutMs`, so an OAuth wait plus two or more slow pages outlives the socket. The client sees a transport timeout, restarts the daemon, resends the listing, and the user gets the second OAuth flow this PR exists to prevent.

Any constant multiple has the same defect — it encodes a phase count the runtime does not promise.

## The fix: observe the daemon instead of predicting it

While a request is in flight the daemon host emits newline-delimited progress frames. The client treats each frame as proof of life and restarts its socket deadline. That turns the socket deadline from an **operation** budget into a **liveness** budget:

- a request survives however many sequential phases it needs — OAuth wait plus N pages — with no arithmetic on either side;
- a daemon that actually goes silent is still torn down and restarted, exactly as before;
- the operation deadline stays where it belongs: the daemon enforces it and answers `operation_timeout`, which the client does not retry.

`resolveOperationSocketBudget` and its 5s grace are gone. The `listTools` socket uses the caller deadline verbatim. Both peers decode frames incrementally (`DaemonFrameDecoder`), so the daemon's own status probe and stop handshake stay readable and a response with no trailing newline still parses. The daemon protocol is versioned, so a daemon started before this change is replaced rather than left speaking the old wire format.

## Real behavior proof

A local OAuth-protected MCP server (dynamic client registration, PKCE `S256`, authorization code, bearer-gated `/mcp`) that paginates `tools/list` across 4 pages at 4s each. `mcporter auth` runs the real interactive browser flow — the browser launch is shimmed so approval lands 4s after the URL opens — routed through the **real keep-alive daemon** (`lifecycle.mode = keep-alive`, confirmed daemon pid). `MCPORTER_OAUTH_TIMEOUT_MS=6000`, so every individual phase is comfortably inside its deadline and only the *total* exceeds a fixed budget.

**Before — 462865b, the head that was closed (`2 * 6000 + 5000` = 17s socket budget):**

```
[14:45:41.300] mcp -> 401 (no valid bearer token), advertising protected resource metadata
[14:45:45.352] authorize -> issuing code, redirecting to loopback callback
[14:45:45.362] token -> PKCE verified, access token issued
[14:45:45.388] tools/list page 1/4
[14:45:49.394] tools/list page 2/4
[14:45:53.398] tools/list page 3/4
[14:45:57.402] tools/list page 4/4
[14:45:58.729] tools/list page 1/4      <-- socket expired at ~17s; daemon restarted, listing replayed
[14:46:02.732] tools/list page 2/4
[14:46:06.735] tools/list page 3/4
[14:46:10.737] tools/list page 4/4

CLI exit 0, elapsed 34s
tools/list page 1 requests : 2
tools/list requests total  : 8
```

**After — this branch:**

```
[14:44:51.258] mcp -> 401 (no valid bearer token), advertising protected resource metadata
[14:44:55.771] authorize -> issuing code, redirecting to loopback callback
[14:44:55.777] token -> PKCE verified, access token issued
[14:44:55.795] tools/list page 1/4
[14:44:59.800] tools/list page 2/4
[14:45:03.805] tools/list page 3/4
[14:45:07.810] tools/list page 4/4

CLI exit 0, elapsed 21s
tools/list page 1 requests : 1
tools/list requests total  : 4
```

One authorization, one completed listing, no replay, no daemon restart. The replay in the "before" run begins 17.3s after the listing started, which is exactly the `2 * timeoutMs + 5s` budget — the failure mode described in the #237 review, reproduced and then removed.

(The provider is a local fixture rather than a third-party service so the run is inspectable and repeatable; it exercises the real SDK OAuth client path — discovery, registration, PKCE, code exchange, bearer-gated MCP — not a stub.)

## Review round: what changed

**[P1] Liveness frames could outlive a wedged request.** The premise as stated — that `listResources`/`readResource` have *no* operation timeout — does not hold: every SDK request carries `DEFAULT_REQUEST_TIMEOUT_MSEC` (60s) and every OAuth wait carries `DEFAULT_OAUTH_CODE_TIMEOUT_MS` (300s). But the underlying worry was real and I found a stronger case than the one reported: `McpRuntime.listTools()` walks `nextCursor` with no guard, so a server that repeats a cursor pages forever — and heartbeats would then keep the client waiting forever. Both are now closed:

- operations with a **fixed** phase count (at most one connect plus one MCP request) carry an absolute ceiling equal to the sum of the deadlines the daemon already applies to those phases. It is derived, not guessed, and can only fire once every real deadline has been blown; a server that accepts and never answers now yields a non-retryable `operation_timeout` instead of an indefinite wait.
- `listTools`, whose page count is data-dependent, keeps its per-page deadline and gains a **repeated-cursor guard** in the runtime.

**[P2] Short deadlines could expire before the first frame.** Fixed, and slightly further than suggested: the first frame is written immediately when the request is dispatched, *and* the cadence is derived from the caller's own deadline (`resolveProgressInterval`, sent on the request envelope) rather than fixed at 250ms. An immediate frame alone would still lose a sub-250ms deadline on the *second* gap.

**[P3] Release-owned changelog.** Correct — checked against this repo's history: my own merged #234 did not touch `CHANGELOG.md`; the maintainer added the entry at release with the `thanks @…` credit line. The entry has been removed from this branch.

**[P1, ClawSweeper] An upgraded client could stop a live shared daemon mid-request.** The protocol-version check classified any pre-v2 daemon as stale and called `stop()` on it from `ensureDaemon()`. A second client upgrading mid-request — OAuth code wait or a delayed cursor page — could race ahead of the first client's in-flight call, kill the daemon, and force the very replay this PR is meant to remove. The replacement is now an idle-drain transition, not a preempt:

- the host's `status` response reports the live in-flight count (`StatusResult.activeRequests`, sampled at response time so a long-running request still shows as in flight). The field is optional, so v1 daemons that cannot advertise it still answer old clients.
- `restartDaemon()` polls the live daemon up to `MCPORTER_DAEMON_DRAIN_TIMEOUT_MS` (60s default, env-var overridable) for the counter to read zero before issuing `stop`. Pre-v2 daemons are treated as permanently busy, so the drain timeout is also their safety bound.
- a busy daemon that another client replaces mid-drain is left alone — `waitForDaemonIdle` returns both `drained` and `replacedByPeer`, and the upgrading client bails out without `stop`.

If the daemon is still busy past the timeout the replacement is **abandoned** and the caller fails loud rather than killing the peer.

## Regression coverage

`tests/daemon-listtools-progress.test.ts` drives the **real** client transport against the **real** host framing over a socket:

- **`survives an OAuth wait followed by several paginated tools/list pages`** — an OAuth wait plus three delayed pages, every phase longer than the socket deadline. Asserts the tools resolve, `listTools` ran exactly **once** (no replay), one `listTools` request reached the daemon, and `launchDaemonDetached` was **never** called (no restart). Verified it fails without the mechanism: stubbing out the progress emitter makes it die with `Daemon did not stop before restart could begin.`
- **`reaches a deadline shorter than the default progress interval`** — the P2 case.
- **`emits the first progress frame before any cadence interval can elapse`** — pins the P2 fix at a 30ms deadline, well under the 250ms default cadence.
- **`still trips the socket deadline when the daemon stops sending progress frames`** — silence is still fatal.
- **`walks every cursor page but refuses to page forever on a repeated cursor`** — the P1 case.

`tests/daemon-client-timeout.test.ts` pins the budget itself: the `listTools` socket deadline equals the caller deadline, so a phase-count multiplier cannot come back unnoticed.

`tests/daemon-host.test.ts` pins the operation ceiling for the fixed-phase methods against a wedged runtime: `callTool`, `listResources`, and `readResource` all return `operation_timeout` after the sum of `MCPORTER_OAUTH_TIMEOUT_MS` and `DEFAULT_REQUEST_TIMEOUT_MSEC`, using fake timers so the test does not wait the production minute-plus.

`tests/daemon-client-config-stale.test.ts` pins the ClawSweeper P1 fix:

- **`waits for a busy shared daemon to drain before sending stop`** — the upgrading client observes `activeRequests` flip to zero before issuing `stop`, proving no in-flight OAuth or cursor page was killed.
- **`does not stop a peer fresh daemon if it replaces the busy one mid-drain`** — a second client that wins the race to replace the busy daemon during the drain window has its fresh daemon left alone; the upgrading client does not call `stop` and does not relaunch.
- **`refuses to replace a daemon that never drains before the timeout`** — `MCPORTER_DAEMON_DRAIN_TIMEOUT_MS=500` exercises the refusal branch without waiting the production minute, asserting the busy peer's daemon is left running.

## Housekeeping

- Clean commit history, no AI co-author trailers, single author.
- Rebased onto current `main`.
- `pnpm check` and `pnpm test` green locally (860 passed, 3 skipped); CI green on Ubuntu, macOS and Windows.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -325,7 +325,7 @@ export async function runCli(argv: string[]): Promise<void> {
         return;
       }
       const { handleAuth: importedHandleAuth } = await import('./cli/auth-command.js');
-      await importedHandleAuth(runtime, resolvedArgs);
+      await importedHandleAuth(runtime, resolvedArgs, { oauthTimeoutMs: runtimeOptionsWithPath.oauthTimeoutMs });
       return;
     }
 
@@ -441,7 +441,7 @@ async function invokeAuthCommand(runtimeOptions: RuntimeOptions, args: string[])
   ]);
   const runtime = await createRuntime(runtimeOptions);
   try {
-    await importedHandleAuth(runtime, args);
+    await importedHandleAuth(runtime, args, { oauthTimeoutMs: runtimeOptions.oauthTimeoutMs });
   } finally {
     await runtime.close().catch(() => {});
   }
@@ -630,6 +630,7 @@ function createDaemonOnlyRuntime(daemonClient: import('./daemon/client.js').Daem
         autoAuthorize: options?.autoAuthorize,
         allowCachedAuth: options?.allowCachedAuth,
         disableOAuth: options?.disableOAuth,
+        timeoutMs: options?.timeoutMs,
       })) as Awaited<ReturnType<Runtime['listTools']>>,
     callTool: (server, toolName, options) =>
       daemonClient.callTool({
--- a/src/cli/auth-command.ts
+++ b/src/cli/auth-command.ts
@@ -4,7 +4,7 @@ import type { OAuthAuthorizationRequest, OAuthSessionOptions } from '../oauth.js
 import { analyzeConnectionError } from '../error-classifier.js';
 import { clearOAuthCaches } from '../oauth-persistence.js';
 import type { createRuntime } from '../runtime.js';
-import { isOAuthFlowError } from '../runtime/oauth.js';
+import { isOAuthFlowError, resolveOAuthTimeoutFromEnv } from '../runtime/oauth.js';
 import type { EphemeralServerSpec } from './adhoc-server.js';
 import { extractEphemeralServerFlags } from './ephemeral-flags.js';
 import { persistPreparedEphemeralServer, prepareEphemeralServerTarget } from './ephemeral-target.js';
@@ -17,12 +17,17 @@ type Runtime = Awaited<ReturnType<typeof createRuntime>>;
 
 type BrowserSuppression = 'default' | 'no-browser';
 
+export interface AuthCommandOptions {
+  readonly oauthTimeoutMs?: number;
+}
+
 const TRUE_VALUES = new Set(['1', 'true', 'yes']);
 const FALSE_VALUES = new Set(['0', 'false', 'no']);
 
-export async function handleAuth(runtime: Runtime, args: string[]): Promise<void> {
+export async function handleAuth(runtime: Runtime, args: string[], options: AuthCommandOptions = {}): Promise<void> {
   const browserSuppression = consumeBrowserSuppression(args, process.env);
   const noBrowser = browserSuppression === 'no-browser';
+  const oauthTimeoutMs = options.oauthTimeoutMs ?? resolveOAuthTimeoutFromEnv();
   let authorizationOutputEmitted = false;
   const markAuthorizationOutputEmitted = () => {
     authorizationOutputEmitted = true;
@@ -85,6 +90,11 @@ export async function handleAuth(runtime: Runtime, args: string[]): Promise<void
       const tools = await withInfoLogsSuppressed(noBrowser, () =>
         runtime.listTools(target, {
           autoAuthorize: true,
+          // The interactive OAuth browser flow runs inside this listTools
+          // request. Let it ride for as long as we're willing to wait for the
+          // authorization code (MCPORTER_OAUTH_TIMEOUT_MS, default 300s) instead
+          // of being killed by the SDK's 60s request cap.
+          timeoutMs: oauthTimeoutMs,
           ...(noBrowser
             ? {
                 oauthSessionOptions: buildNoBrowserOAuthOptions(format, markAuthorizationOutputEmitted),
--- a/src/daemon/client.ts
+++ b/src/daemon/client.ts
@@ -163,15 +163,16 @@ export class DaemonClient {
       // would silently return null on a pid mismatch and the busy-daemon
       // regression would slip past us.
       const liveStatus = await this.probeLiveStatus();
-      if (
-        liveStatus &&
-        options.expectedPid !== undefined &&
-        liveStatus.pid !== options.expectedPid &&
-        (await this.checkConfigState()) === 'fresh'
-      ) {
+      if (options.reason === 'stale-config' && liveStatus && (await this.checkConfigState()) === 'fresh') {
         return;
       }
-      if (options.reason === 'stale-config' && liveStatus && (await this.checkConfigState()) === 'fresh') {
+      // If the daemon has already been replaced by another client (PID
+      // mismatch before the drain even starts), the replacement is theirs to
+      // own -- stopping it would kill their in-flight request. The config
+      // check above is not enough: the peer may not have rewritten the
+      // metadata yet, so `checkConfigState` can still report `stale` and the
+      // condition would fall through to `stop()`.
+      if (liveStatus && options.expectedPid !== undefined && liveStatus.pid !== options.expectedPid) {
         return;
       }
       // Codex / ClawSweeper P1: an upgraded client must not stop a daemon that
--- a/src/daemon/client.ts
+++ b/src/daemon/client.ts
@@ -158,16 +158,20 @@ export class DaemonClient {
 
   private async restartDaemon(options: { reason?: 'stale-config'; expectedPid?: number } = {}): Promise<void> {
     await this.startingWithLock(async () => {
-      const currentStatus = await this.readVerifiedStatus();
+      // Probe the daemon without the pid-match filter so the drain check
+      // below can see whether the daemon is actually busy. `readVerifiedStatus`
+      // would silently return null on a pid mismatch and the busy-daemon
+      // regression would slip past us.
+      const liveStatus = await this.probeLiveStatus();
       if (
-        currentStatus &&
+        liveStatus &&
         options.expectedPid !== undefined &&
-        currentStatus.pid !== options.expectedPid &&
+        liveStatus.pid !== options.expectedPid &&
         (await this.checkConfigState()) === 'fresh'
       ) {
         return;
       }
-      if (options.reason === 'stale-config' && currentStatus && (await this.checkConfigState()) === 'fresh') {
+      if (options.reason === 'stale-config' && liveStatus && (await this.checkConfigState()) === 'fresh') {
         return;
       }
       // Codex / ClawSweeper P1: an upgraded client must not stop a daemon that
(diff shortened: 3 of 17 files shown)


umutkeltek:
Thanks — both review passes found real things. All three items are addressed and the PR body now carries an inspectable before/after run. Summary:

**P1 — liveness frames outliving a wedged request.** The stated premise doesn't hold: `listResources`/`readResource` do have an operation timeout, because every SDK request carries `DEFAULT_REQUEST_TIMEOUT_MSEC` (60s, `shared/protocol.js`) and every OAuth wait carries `DEFAULT_OAUTH_CODE_TIMEOUT_MS` (300s). But chasing it turned up a *stronger* case than the one reported: `McpRuntime.listTools()` walks `nextCursor` with no guard, so a server that repeats a cursor pages forever — and heartbeats would then keep a caller waiting forever. Both ends are now closed:

- fixed-phase operations (at most one connect plus one MCP request) carry an absolute ceiling equal to the **sum of the deadlines the daemon already applies** to those phases — derived rather than guessed, so it can only fire once every real deadline has been blown. A server that accepts and never answers now returns a non-retryable `operation_timeout` instead of hanging.
- `listTools` — the one operation whose phase count is data-dependent — keeps its per-page deadline and gains a repeated-cursor guard.

**P2 — short deadlines expiring before the first frame.** Fixed, and a step further than suggested: an immediate first frame alone still loses a sub-250ms deadline on the *second* gap, so the cadence is now derived from the caller's own deadline (sent on the request envelope) **and** the first frame is written immediately on dispatch.

**P3 — release-owned changelog.** Correct, and I checked it against this repo's history rather than taking it on trust: my own merged #234 didn't touch `CHANGELOG.md` — the maintainer added the entry at release with the `thanks @…` credit. Entry removed.

**Real behavior proof** is in the PR body: a local OAuth-protected MCP server (dynamic registration, PKCE `S256`, bearer-gated `/mcp`) paginating `tools/list` over 4 pages, driven by a real `mcporter auth` browser flow through the real keep-alive daemon. On 462865b the listing is replayed — page 1 requested twice, 8 `tools/list` requests, 34s — with the replay starting 17.3s in, exactly the `2 * timeoutMs + 5s` budget. On this branch: one authorization, 4 requests, 21s, no restart, no replay.

Each item has regression coverage; `pnpm check` and `pnpm test` are green locally (851 passed) and CI is green on Ubuntu, macOS and Windows.

umutkeltek:
Fixed the remaining P2 — good catch, the finding is exact.

`resolveProgressInterval()` clamped the cadence up to a 25ms floor, which meant that for any caller deadline at or below 25ms the interval overshot the very deadline it was meant to refresh: the immediate frame landed, then the socket expired before the next one, and `invoke()` classified that as a dead transport and replayed. The floor was the one constant left in a mechanism whose whole point is not to use constants, so it is gone rather than lowered — the cadence is now `min(250, max(1, floor(deadline / 3)))`, strictly inside the caller's deadline for every deadline above 1ms.

I did not reject short values at the input boundary instead: `timeoutMs: 1` is accepted today and pinned by an existing test (`clamps daemon status preflight timeout for tiny per-call timeouts`), so tightening the boundary would be a separate behavior change. A 1ms deadline stays unachievable at any cadence — as it was before progress frames existed.

Regression added below the old floor, asserting the invariant directly rather than racing a timer: `resolveProgressInterval(n) < n` across 2, 3, 5, 12, 24, 25, 74, 75, 300, 30_000, 300_000, plus the 1ms boundary and the default-cadence fallbacks.

`pnpm check` and `pnpm test` green (853 passed, 3 skipped).

steipete:
Heads-up for the next rebase: #253 just landed on main and overlaps this PR's territory. `shouldRestartDaemonServer` in `src/daemon/runtime-wrapper.ts` now returns false for unauthorized errors (restart/replay on 401 was one source of the duplicate OAuth prompts in #247), and `tests/keep-alive-runtime.test.ts` grew a case locking that in.

That may shrink what this PR needs to solve — the OAuth-replay motivation is partly addressed — but the core liveness-frame design for multi-phase `listTools` deadlines is still open and worth pursuing. The remaining review item is the P1 active-request accounting defect (a status probe counting itself as active work, preventing idle drain); once that's fixed and the branch is rebased over #253/#248, this is ready for another look.

steipete:
Triage update, and an apology for the timing.

**Your analysis holds, and the bug is still real on `main`.** I verified it against 0.13.0: `listTools` reaches `invoke('listTools', params)` with no timeout, so it falls back to the flat `DEFAULT_DAEMON_TIMEOUT_MS` of 30s, while the daemon-side handler calls `runtime.listTools(...)` with `autoAuthorize` and will wait up to `DEFAULT_OAUTH_CODE_TIMEOUT_MS` — 5 minutes — for browser consent. There are no liveness frames in the daemon protocol. So a user who spends more than 30 seconds at the consent screen still gets a client-side transport timeout, a daemon restart, and the second OAuth prompt you set out to eliminate.

Your central point survives too: `resolveOperationSocketBudget` is gone, but what replaced it is a flat 30s, which is the same mistake in simpler clothing — a deadline that encodes an assumption about how long legitimate work takes. The liveness-budget framing is the right answer, and "observe the daemon instead of predicting it" is a better articulation of it than anything in the current code.

**The bad news is that this branch cannot be rebased.** Between your submission and now, the repo migrated to MCP SDK v2 for the 2026-07-28 protocol (#255), reworked elicitation and the daemon runtime (#256, #262), and rewrote large parts of `src/daemon/{client,host,protocol}.ts`, `src/runtime.ts`, `src/cli.ts` and `src/cli/auth-command.ts` — 17 of the files your patch touches. GitHub already marks it CONFLICTING, and the conflict is semantic rather than textual: the manual `tools/list` cursor walk your budget arithmetic reasoned about was replaced by the SDK's aggregating walk.

Asking you to redo 1,500 lines against a moved target would be a poor trade for your time, so I am reimplementing your design directly, with `Co-authored-by: Umut Keltek <umut.keltek@gmail.com>` on the commits and a changelog credit. The frame decoder, the progress-frame liveness reset, the daemon-side operation deadline with a non-retryable timeout code, and the protocol version for client/daemon skew all come from your patch — I am adapting them to the current shapes, not replacing the design.

I will link the replacement PR here and close this one once it lands. If you would rather carry it yourself against 0.13.0, say so and I will stand down — it is your work and I would rather you got the merge.

Thank you for the report and for the second iteration on it; the diagnosis was correct and it outlived the code it was written against.

steipete:
Landed as #270 (`5ea1970`), with your `Co-authored-by` on the commit and a changelog credit under 0.13.1.

The implementation follows your design: progress frames from the daemon, an idle deadline on the client that they refresh, a daemon-owned operation deadline returned as a non-retryable error so a slow request is never replayed into a second prompt, and incremental frame decoding so split and coalesced writes parse. Version skew is handled in both directions — progress emission requires an explicit v2 opt-in, so an old client never sees a frame it cannot parse, and a new client talking to a daemon that outlived its build simply retains the flat deadline.

One detail from verification you may find interesting: Node's `socket.setTimeout` is an inactivity timer, so incoming progress data resets it natively. The explicit re-arm on the client is redundant belt-and-braces — the real mechanism is the daemon writing at all. I confirmed that by mutation: suppressing the daemon's emission fails the liveness test with `Daemon request timed out`, while deleting the client-side re-arm changes nothing. Worth knowing before someone "simplifies" that line away.

Closing this in favour of the replacement, but the diagnosis and the design are yours — thank you for both, and for the second iteration on the original report. Sorry the migration landed on top of it.

umutkeltek:
Thank you. This is a great outcome. I am happy for #270 to be the canonical implementation, and I really appreciate you preserving the design, carrying the attribution through as co-author and in the changelog, and verifying the original failure mode against current main.

The socket.setTimeout inactivity behavior is a useful detail too. It is good to know that the daemon’s progress writes are the actual mechanism and the explicit re-arm is just belt and braces.

No apology needed. The migration timing was simply unlucky. I am glad the diagnosis and liveness-budget framing survived the rewrite and landed cleanly. Thanks again for handling this so thoughtfully.

PR #246, opened 2026-07-28
test(oauth): cover credential-generation compare-and-set helpers

KrasimirKralev:
## What Problem This Solves

`src/oauth-token-generation.ts` exports eight helpers that stamp each OAuth credential save with an opaque **generation** marker, so the vault/persistence layer can do compare-and-set writes across multiple backing stores (`oauth-vault.ts:167/173/214/308/320/363`, `oauth-persistence.ts`). They decide which concurrent store write *wins* — yet shipped with **no direct test coverage** (they are only exercised transitively through save/load round-trips; a repo-wide search finds no test importing the module).

The contract is subtle and easy to regress silently:

- `withOAuthTokenGeneration` / `withOAuthClientGeneration` **reuse** an existing generation and only **mint** a new one (`randomUUID`) when none is present — so two independent stamps of the same value are *distinct* saves;
- `withHiddenOAuthTokenGeneration` / `withHiddenOAuthClientGeneration` make the marker **non-enumerable** (hidden from JSON/API values) while keeping it readable for recovery, and return the input **unchanged** when there is nothing to hide;
- `sameOAuthTokenGeneration` / `sameOAuthClientGeneration` require **matching generation AND matching public value** when a generation is present, return `false` for an absent current, and fall back to **deep-equality** only for two legacy (generation-less) values — so a generationed value never matches a legacy one;
- `sameOAuthTokenValue` / `sameOAuthClientValue` compare public values while ignoring *only* the internal marker.

A regression in any of these — dropping the reuse, leaking the marker into serialized values, or dropping the public-value check from the compare-and-set — would let a stale write win or a live write be rejected, and would not be caught before merge.

## Why This Change Was Made

Coverage-only. This adds one new file, `tests/oauth-token-generation.test.ts` (**+115, no production code touched**), pinning all eight exported helpers' current `main` behavior. The tests drive the **real exported functions** directly (no stub), asserting enumerability via `Object.keys`/`JSON` round-trips and identity via reference checks, so they exercise the production compare-and-set path the persistence layer relies on. No new config, defaults, dependencies, or behavior — the diff is a single test file.

## User Impact

No user-visible or runtime change. For maintainers, the credential-generation contract now regresses loudly instead of silently: a future edit that breaks generation reuse, the hidden-marker semantics, or the generation-plus-value comparison will fail this suite.

## Evidence

Linux, Node 22.22, `pnpm install --frozen-lockfile` from source, branched off current `main` (`7259c8c`).

**13/13 pass on clean `main`:**

```text
$ node_modules/.bin/vitest run tests/oauth-token-generation.test.ts
 Test Files  1 passed (1)
      Tests  13 passed (13)
```

**Non-vacuous — the suite bites when the target is mutated** (each mutation applied to `oauth-token-generation.ts`, suite re-run, then reverted byte-identical):

| Mutation to `oauth-token-generation.ts` | Result |
| --- | --- |
| `withOAuthTokenGeneration` always mints (drop the reuse of an existing generation) | 1 fail |
| `sameOAuthTokenValue` compares full objects (stop ignoring the generation) | 4 fail |
| hidden marker made enumerable (`enumerable: true`) | 1 fail |
| `sameOAuthTokenGeneration` drops the public-value check | 1 fail |
| *(reverted — control)* | **13 pass** |

**Format / lint / types:**

```text
$ node_modules/.bin/oxfmt --check tests/oauth-token-generation.test.ts   # All matched files use the correct format.
$ node_modules/.bin/oxlint --type-aware --tsconfig tsconfig.json --deny-warnings tests/oauth-token-generation.test.ts   # exit 0
$ node_modules/.bin/tsc --project tsconfig.json --noEmit   # exit 0
```

**Scope note:** one new `*.test.ts` under `tests/`, `+115 / -0`, no production code touched. The randomness in `randomUUID` is only exercised through the deterministic distinct-vs-reuse contract (two stamps differ; a re-stamp preserves), so the suite is stable.

**Related:** none — coverage mined from the module itself; no linked issue.

---

_AI-assisted contribution._

---
_Generated by [Claude Code](https://claude.ai/code/session_01C3Frhg6fZrLpLXB6GGDpvx)_

### DIFF
--- a/tests/oauth-token-generation.test.ts
+++ b/tests/oauth-token-generation.test.ts
@@ -1,0 +1,115 @@
+import type { OAuthClientInformationMixed, OAuthTokens } from '@modelcontextprotocol/sdk/shared/auth.js';
+import { describe, expect, it } from 'vitest';
+import {
+  sameOAuthClientGeneration,
+  sameOAuthClientValue,
+  sameOAuthTokenGeneration,
+  sameOAuthTokenValue,
+  withHiddenOAuthClientGeneration,
+  withHiddenOAuthTokenGeneration,
+  withOAuthClientGeneration,
+  withOAuthTokenGeneration,
+} from '../src/oauth-token-generation.js';
+
+// These live helpers stamp each credential save with an opaque "generation" so
+// oauth-vault.ts / oauth-persistence.ts can compare-and-set across backing stores,
+// yet ship with no direct unit test. Pins reuse-vs-mint, the hidden (non-enumerable)
+// marker, and the generation-plus-value / legacy-deep-equal comparison arms.
+
+const baseTokens = { access_token: 'tok-abc', token_type: 'Bearer' } as OAuthTokens;
+const baseClient = { client_id: 'client-xyz' } as OAuthClientInformationMixed;
+
+describe('withOAuthTokenGeneration', () => {
+  it('stamps a new enumerable generation on tokens that lack one, without mutating the input', () => {
+    const stamped = withOAuthTokenGeneration(baseTokens);
+    expect(sameOAuthTokenValue(stamped, baseTokens)).toBe(true); // public value preserved
+    expect(Object.keys(stamped)).toHaveLength(Object.keys(baseTokens).length + 1); // extra enumerable marker
+    expect(Object.keys(baseTokens)).toEqual(['access_token', 'token_type']); // input untouched
+  });
+
+  it('reuses an existing generation instead of minting a new one', () => {
+    const stamped = withOAuthTokenGeneration(baseTokens);
+    const restamped = withOAuthTokenGeneration(stamped);
+    expect(sameOAuthTokenGeneration(restamped, stamped)).toBe(true);
+  });
+
+  it('mints distinct generations across independent stamps of the same value', () => {
+    const a = withOAuthTokenGeneration(baseTokens);
+    const b = withOAuthTokenGeneration(baseTokens);
+    expect(sameOAuthTokenValue(a, b)).toBe(true); // identical public value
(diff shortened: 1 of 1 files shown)


steipete:
Maintainer repair and verification are complete at `ea0981fb6de7d8f7a628b38b0684f668ba427bef`.

I added the two missing client-side compare-and-set cases from review: restamping preserves an existing client generation, and a changed public client value is rejected even when it carries the same generation. I also removed the redundant assertion-restatement comments. The contributor's original commit remains intact.

Proof on Node 26.1.0 / pnpm 10.33.2:

- `pnpm exec vitest run tests/oauth-token-generation.test.ts` — 15 passed.
- `pnpm check` — formatting, type-aware lint, and typecheck passed.
- `pnpm test` — 131 files passed, 1 skipped; 855 tests passed, 3 skipped.
- `pnpm build` — passed.
- Direct compiled-library run — `{"reuseAccepted":true,"staleRejected":true}`.
- AutoReview (Codex, branch versus `origin/main`) — clean, no accepted/actionable findings.
- Public model-identifier gate — PASS; this test-only diff has no model-bearing surfaces.

Maintainer disposition: LAND once exact-head CI completes successfully.

steipete:
Landed as ced43e458a8bc8d7305cda79bbd5ae7025cd2220.

Verification at exact head `ea0981f` before merge:

- `pnpm exec vitest run tests/oauth-token-generation.test.ts` — 15 passed (local, Node 24.18.0 / pnpm 10.33.2).
- Cross-platform CI builds (ubuntu / macos-15 / windows) all green on the head.
- Full-suite, build, and Codex AutoReview proof was already recorded in the maintainer verification comment above and the head has not changed since.

Test-only change, so no changelog entry. Thanks @KrasimirKralev!

PR #251, opened 2026-08-02
docs(known-issues): mark SDK inline-stdio Status line as stale

LDMB123:
The "MCP SDK 1.22.0 inline-stdio regression" entry's Status line still says the SDK is "pinned to `~1.21.2` until upstream ships a fix" — but `package.json` has carried `@modelcontextprotocol/sdk: ^1.30.0` for a while now, so the documented remediation no longer exists.

This PR only corrects the Status line to say what is actually known: the pin is gone, and whether the underlying `typeName` regression was fixed upstream, worked around in the inline-stdio test harness, or is simply no longer exercised has not been re-verified. It deliberately does not claim the regression is fixed — it points readers at the existing repro to re-verify before relying on the entry.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


steipete:
Landed as b88f4a2bb4b36c443a6796a398948d532c6e53d6.

Verification before merge:

- Confirmed the stale claim: `origin/main` `package.json` carries `@modelcontextprotocol/sdk` `^1.30.0`, so the documented `~1.21.2` pin no longer exists. The rewritten Status line accurately says the regression state is unverified rather than fixed.
- Build matrix (ubuntu / macos-15 / windows) green on the head.
- The one failing `dispatch` check was ClawSweeper's dispatch job hitting its own GitHub API installation rate limit (HTTP 403) — unrelated to this change; the review ran later regardless.

Docs-only, no changelog entry. Thanks @LDMB123!

PR #287, opened 2026-08-08
fix(vault): accept dynamic client registration clientInfo in vault set

### LINKED ISSUE #286: vault set rejects clientInfo from OAuth dynamic client registration
## What happens

`mcporter vault set <server> --stdin` rejects client info that came from OAuth dynamic client registration. It prints:

```
[mcporter] Vault payload clientInfo.redirect_uris must be a string.
```

`redirect_uris` is an array, not a string. RFC 7591 registration responses always return it as an array. So any real DCR result is rejected and nothing is saved.

This worked in 0.12.0. It broke in 0.12.1 and is still broken in 0.13.0.

## Steps to reproduce

1. Create `~/.mcporter/mcporter.json`:

   ```json
   {"mcpServers":{"demo":{"url":"https://example.test/mcp","auth":"oauth"}}}
   ```

2. Pipe a payload to `vault set`:

   ```bash
   echo '{"tokens":{"access_token":"fake","token_type":"Bearer"},"clientInfo":{"client_id":"abc","redirect_uris":["https://example.test/cb"],"grant_types":["authorization_code"],"response_types":["code"],"token_endpoint_auth_method":"none"}}' \
     | mcporter vault set demo --stdin
   ```

The access token is fake, so no real credential is involved. The server URL is never contacted.

### What you get

On 0.12.1 and later, it exits with code 1 and prints:

```
[mcporter] Vault payload clientInfo.redirect_uris must be a string.
```

On 0.12.0, the same command succeeds:

```
Saved OAuth credentials for 'demo' to <data-dir>/mcporter/credentials.json
```

If you run the 0.12.0 case, remove the test entry afterwards with `mcporter vault clear demo`.

### What I expected

The payload is saved. `redirect_uris` is a normal part of client info.

## Where it comes from

In `dist/cli/vault-command.js`:

```js
function validateOAuthClientInfo(clientInfo) {
    for (const [key, value] of Object.entries(clientInfo)) {
        if (value !== undefined && value !== null && typeof value !== 'string') {
            throw new CliUsageError(`Vault payload clientInfo.${key} must be a string.`);
        }
    }
}
```

This requires every value to be a string. But `dist/oauth-client-info.d.ts` types client info as `OAuthClientInformationMixed` from the MCP SDK, and in `@modelcontextprotocol/sdk/shared/auth` that is:

```ts
type OAuthClientInformationMixed = OAuthClientInformation | OAuthClientInformationFull;
```

Neither member of that union can pass the check:

- `OAuthClientInformation` has `client_id_issued_at` and `client_secret_expires_at`, both numbers.
- `OAuthClientInformationFull` adds `redirect_uris`, `grant_types` and `response_types`, all arrays of strings.

So the only client info the command accepts is `client_id` plus optional string fields. Any registration response that carries issuance timestamps or metadata is refused, even though the declared type allows it.

`--tokens-file` has the same problem. Both sources go through `validateVaultPayload`, so there is no way around it.

## Versions

| Version | Result |
| --- | --- |
| 0.12.0 | works |
| 0.12.1 | fails |
| 0.12.2 | fails |
| 0.12.3 | fails |
| 0.12.4 | fails |
| 0.13.0 | fails |

Node 24, macOS. Nothing here looks platform specific.

## Suggested fix

Check each field for its own type instead of requiring strings everywhere:

- `redirect_uris`, `grant_types`, `response_types` — array of strings
- `client_id_issued_at`, `client_secret_expires_at` — number
- `client_id`, `client_secret`, `client_name`, `token_endpoint_auth_method` — string

Dropping the `clientInfo` check and keeping the `tokens` checks would also fix it.

## Related

#244 is also about `clientInfo` on vault entries, but the other way around: it goes missing after authorization. These may be worth looking at together.

## Workaround

For anyone blocked on this: the write goes through if you trim `clientInfo` to the string-only subset the validator accepts before piping it to `vault set`.

With `jq`:

```bash
jq -c '.clientInfo |= with_entries(select(.value | type == "string"))' payload.json \
  | mcporter vault set demo --stdin
```

Applied to the payload from the report, that leaves `{"client_id":"abc","token_endpoint_auth_method":"none"}` and the command succeeds on 0.12.1+.

The same trim in TypeScript, if you are shelling out to `vault set` from code:

```ts
/** The string-only subset of client info that `mcporter vault set` accepts. */
interface VaultClientInfo {
  client_id: string;
  token_endpoint_auth_method: string;
  client_secret?: string;
  client_name?: string;
}

function toVaultClientInfo(info: OAuthClientInformationFull): VaultClientInfo {
  return {
    client_id: info.client_id,
    token_endpoint_auth_method: info.token_endpoint_auth_method,
    ...(info.client_secret !== undefined ? { client_secret: info.client_secret } : {}),
    ...(info.client_name !== undefined ? { client_name: info.client_name } : {}),
  };
}
```

## What it costs

Dropping `redirect_uris`, `grant_types`, `response_types`, `client_id_issued_at` and `client_secret_expires_at` is only safe when the caller owns registration:

- Refreshes keep working. A `refresh_token` grant sends no redirect URI, and the MCP SDK reads `client_id` and `token_endpoint_auth_method` when it authenticates the refresh call — so those two are the ones worth keeping.
- If you depend on mcporter re-registering the client, the trimmed entry no longer carries the metadata for it. We do registration ourselves and hand mcporter the result, so mcporter never re-registers.

This is a workaround, not a fix. The validator still contradicts the `OAuthClientInformationMixed` type the command declares it accepts, and a schema-aware validator would let the full registration response through.

Yigtwxx:
## Summary

Fixes #286. `mcporter vault set` rejects every `clientInfo` value that is not a string, so no real dynamic client registration response can be seeded into the vault. In RFC 7591, `redirect_uris`, `grant_types`, `response_types` and `contacts` are string arrays, and `client_id_issued_at` and `client_secret_expires_at` are numbers. A headless deployment that already holds OAuth credentials has no way to install them.

The declared type says otherwise. `VaultPayload.clientInfo` is `OAuthClientInformationMixed`, and mcporter builds that same array-valued shape itself in `buildStaticClientInformation` (`src/oauth-client-info.ts:15`) and during registration (`src/oauth.ts:115`), then reads `clientInfo.redirect_uris` back as an array (`src/oauth.ts:506`). Its own client information cannot round-trip through its own command.

### Root cause

`src/cli/vault-command.ts:154`

```ts
function validateOAuthClientInfo(clientInfo: Record<string, unknown>): void {
  for (const [key, value] of Object.entries(clientInfo)) {
    if (value !== undefined && value !== null && typeof value !== 'string') {
      throw new CliUsageError(`Vault payload clientInfo.${key} must be a string.`);
    }
  }
}
```

One rule for every field. `validateOAuthTokens`, directly above it, already does the opposite: `access_token` and `token_type` non-empty strings, `refresh_token`/`scope`/`issuer` optional strings, `expires_in` a finite number. So the tokens half of the payload accepts its real shape and the client-information half does not.

## Fix

`validateOAuthClientInfo` now checks each field against the JSON type `OAuthClientInformationFullSchema` declares for it (`@modelcontextprotocol/sdk/shared/auth.js`), in that schema's declaration order:

| Type | Fields |
| --- | --- |
| non-empty string, required | `client_id` |
| string | `token_endpoint_auth_method`, `client_name`, `client_uri`, `logo_uri`, `scope`, `tos_uri`, `policy_uri`, `jwks_uri`, `software_id`, `software_version`, `software_statement`, `client_secret`, `issuer` |
| array of strings | `redirect_uris`, `grant_types`, `response_types`, `contacts` |
| finite number | `client_id_issued_at`, `client_secret_expires_at` |

The loop walks the rule table rather than the payload. Unknown keys therefore pass through untouched, so a full registration response keeps `registration_client_uri`, `registration_access_token` and whatever else the provider returned, and the reported field no longer depends on JSON key order. `jwks` stays unconstrained because the SDK types it as `z.any()`. `redirect_uris` stays optional even though the schema's wide member requires it, since the narrow member `OAuthClientInformation` is equally valid here. `null` is still treated as absent, as before.

URL-shaped fields are only checked for being strings. The SDK's `SafeUrlSchema` would additionally require a parseable URL. That starts rejecting redirect URIs that are legal in the wild, such as `urn:ietf:wg:oauth:2.0:oob` and custom-scheme mobile redirects, which is the wrong trade in a change whose point is to stop over-rejecting.

I kept the SDK's zod schemas out of this path for a related reason. They are all `.strip()`, so validating through them invites persisting `result.data`, which would quietly drop seeded fields the vault depends on: `tokens.expires_at` is read at `src/oauth-persistence-stores.ts:37`, `src/oauth-token-refresh.ts:33` and `src/runtime/oauth.ts:123`, and `validateVaultPayload` deliberately returns the raw objects by reference. A union of `OAuthClientInformationFullSchema` and `OAuthClientInformationSchema` would not validate much either, since the second member requires only `client_id` and strips the rest, so `{ client_id: "abc", redirect_uris: 42 }` passes it.

`validateVaultPayload` is unchanged, including the order in which it validates tokens before client information.

### Also: three fields the read path requires and the write path did not check

`externalVaultEntry` (`src/oauth-vault.ts:220`) runs stored credentials through `isStoredOAuthTokens` and `isStoredOAuthClientInformation`, and drops the whole `tokens` or `clientInfo` object when a guard fails. Those guards require `clientInfo.issuer` to be a string and `tokens.expires_at`/`tokens.expiresAt` to be finite numbers. `vault set` checked none of the three, so a typo in one field silently discarded everything next to it, and the command still printed `Saved OAuth credentials`.

`clientInfo.issuer` was covered incidentally by the old string-only rule, so the per-field table has to keep covering it. It is worth naming because `vault set` is the only writer of that field and its only consumer is a security check: `assertRefreshIssuerBinding` (`src/oauth-token-refresh.ts:86`) pins refresh to the recorded issuer, and an unusable value turns the pin into a no-op. `tokens.expires_at` and `expiresAt` are the pre-existing half, folded into the same loop as `expires_in`.

### Deliberate tightening: client_id is now required

Today `{"clientInfo": {}}` is accepted. Both members of `OAuthClientInformationMixed` require `client_id`, so the existing `as OAuthClientInformationMixed` cast does not hold for that value. And `{}` is not read downstream as "no client information": it is a truthy object that `saveVaultEntry` persists and `getClientInformation()` later hands back in place of the statically configured client, so the failure surfaces much later as an opaque SDK error during refresh. `clientIdFromEntry` (`src/oauth-vault.ts:302`) and same-URL credential inheritance both need it too. This does make the write side stricter than `isStoredOAuthClientInformation`, which does not require `client_id`, and it is a behavior change, so I would rather flag it than bury it. It is one `if` plus one test case if you would rather not take it.

### Two adjacent fixes in the same input path

- `JSON.parse` at line 47 was unguarded, so an unparsable payload left the command as a raw `SyntaxError` and printed a stack trace through the unexpected-error path instead of the usage path every other payload problem already uses. It now fails as a `CliUsageError` naming the source. The message deliberately does not repeat the parser's reason: V8 quotes a prefix of the input (`Unexpected token 's', "sk-live-9f"... is not valid JSON`), and the input here is credential material that can end up in a CI log. Naming the file follows `src/cli/call-arguments.ts:373`.
- The `vault set` help text showed `clientInfo` as `client_id` alone, which reads as the whole contract and is plausibly where the expectation in the issue came from. It now says a full registration response is accepted.

## Behavior proof

The real CLI through `tsx src/cli.ts`, throwaway `XDG_CONFIG_HOME` and `XDG_DATA_HOME`, and the payload from the issue verbatim. The access token is fake and `https://example.test/mcp` is never contacted.

```
=== BEFORE (main @ 764bb87) ===
--- mcporter vault set demo --stdin
[mcporter] Vault payload clientInfo.redirect_uris must be a string.
--- stored vault entry
(no credentials file written)
--- mcporter vault set demo --stdin  (truncated JSON)
[mcporter] Unexpected end of JSON input
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)

=== AFTER (this branch) ===
--- mcporter vault set demo --stdin
Saved OAuth credentials for 'demo' to <temp>/mcporter/credentials.json
--- stored vault entry
{
  "tokens": {
    "access_token": "fake",
    "token_type": "Bearer",
    "__mcporter_generation": "a344f320-3658-4b95-b6b4-972186d02740"
  },
  "clientInfo": {
    "client_id": "abc",
    "redirect_uris": [
      "https://example.test/cb"
    ],
    "grant_types": [
      "authorization_code"
    ],
    "response_types": [
      "code"
    ],
    "token_endpoint_auth_method": "none",
    "__mcporter_client_generation": "8468dd59-b501-47ff-a77e-7a76f3b471ab"
  }
}
--- mcporter vault set demo --stdin  (truncated JSON)
[mcporter] Vault payload from stdin is not valid JSON.
```

Same harness for the guard mismatch, reading the entry back through `loadVaultEntry` so the read-path guards actually run:

```
=== BEFORE (main @ 764bb87) ===
--- tokens.expires_at = "soon"
Saved OAuth credentials for 'demo' to <temp>/mcporter/credentials.json
    loadVaultEntry: (entry carries neither tokens nor clientInfo)
--- clientInfo.issuer = 42
[mcporter] Vault payload clientInfo.issuer must be a string.
    loadVaultEntry: (no vault entry)

=== AFTER (this branch) ===
--- tokens.expires_at = "soon"
[mcporter] Vault payload tokens.expires_at must be a finite number.
    loadVaultEntry: (no vault entry)
--- clientInfo.issuer = 42
[mcporter] Vault payload clientInfo.issuer must be a string.
    loadVaultEntry: (no vault entry)
```

The first case is the silent loss: the command reports success and the credentials are unreadable. The second is the one the old rule already caught and the new table has to keep catching.

## Tests

`tests/vault-validation.test.ts` goes from 16 to 32 cases, in the table-driven style the file already uses:

- `accepts a dynamic client registration clientInfo payload` — the issue's payload verbatim.
- `accepts registration timestamps, contacts, and a jwks document` — `client_secret`, `client_id_issued_at`, `client_secret_expires_at: 0`, `contacts`, and a `jwks` document. Both acceptance cases assert the stored entry through `loadVaultEntry`, not the input object, so a dropped `clientInfo` fails the test.
- `preserves unknown clientInfo fields and seeded token expiry` — `issuer`, `registration_client_uri`, `registration_access_token`, `tokens.expires_at` and `id_token` all reach the vault verbatim. This is the case that fails if anyone later swaps the validator for one that reparses the payload.
- `persists null clientInfo fields verbatim` — the pre-0.12.1 null tolerance, held in place.
- `rejects malformed clientInfo` — a non-string entry inside `redirect_uris`, `redirect_uris` and `grant_types` given as bare strings, `client_id_issued_at` and `client_secret_expires_at` given as strings, a non-string `client_name`, a non-string `issuer`, and `clientInfo` with no `client_id`.
- `rejects malformed payload` — the existing rows plus `tokens.expires_at` and `tokens.expiresAt` given as strings.
- `reports malformed stdin JSON as a usage error without echoing the payload` and `names the file when its JSON is malformed` — one per input source.

Reverting only `src/cli/vault-command.ts` to main and keeping the tests:

```
=== tests against the unpatched validator ===
 FAIL  rejects malformed payload 6            (tokens.expires_at)
 FAIL  rejects malformed payload 7            (tokens.expiresAt)
 FAIL  rejects malformed payload 9            (clientInfo.client_id message)
 FAIL  reports malformed stdin JSON as a usage error without echoing the payload
 FAIL  names the file when its JSON is malformed
 FAIL  accepts a dynamic client registration clientInfo payload
 FAIL  accepts registration timestamps, contacts, and a jwks document
 FAIL  rejects malformed clientInfo 0         (redirect_uris entry)
 FAIL  rejects malformed clientInfo 1         (redirect_uris as string)
 FAIL  rejects malformed clientInfo 2         (grant_types as string)
 FAIL  rejects malformed clientInfo 3         (client_id_issued_at)
 FAIL  rejects malformed clientInfo 4         (client_secret_expires_at)
 FAIL  rejects malformed clientInfo 7         (missing client_id)
      Tests  13 failed | 19 passed (32)
```

The other 19 pass on both sides and cover what this change must not alter, including `rejects malformed clientInfo 6`, the `issuer` case.

One committed expectation moved: the row asserting `clientInfo.client_id must be a string` for `client_id: 42` now expects `must be a non-empty string`, matching `tokens.access_token`.

## Gates

Windows 11, Node 22.20.0, pnpm 10.33.2.

- `pnpm exec vitest run tests/vault-validation.test.ts tests/vault-command.test.ts` — 37 passed.
- `pnpm check` — oxfmt, type-aware oxlint and `tsc --noEmit` clean.
- `pnpm build` — clean.
- `pnpm test` — 1283 passed, 84 skipped, 7 failed. The 7 are in `tests/chrome-devtools-relay-handoff.test.ts` and `tests/runtime-chrome-relay-handoff.test.ts`, and they fail identically on unmodified `main` at 764bb87 on this machine (`Invalid URL` from the relay handoff), so they predate this branch and touch nothing it changes. I could not clear them here because this box is below the engine floor the repo declares (`node >=24`). CI settles it: `build` passes on ubuntu-latest, macos-15 and windows-latest for this branch.

Thanks @feniix for the report. The version table and the pointer straight at the validator left nothing to guess.

### DIFF
--- a/src/cli/vault-command.ts
+++ b/src/cli/vault-command.ts
@@ -9,6 +9,8 @@ interface VaultPayload {
   readonly clientInfo?: OAuthClientInformationMixed;
 }
 
+type VaultPayloadSource = { kind: 'file'; path: string } | { kind: 'stdin' };
+
 export interface VaultCommandOptions {
   readonly readStdin?: () => Promise<string>;
 }
@@ -44,7 +46,7 @@ async function handleVaultSet(
     throw new CliUsageError(`Unknown vault set argument '${args[0]}'.`);
   }
   const definition = runtime.getDefinition(server);
-  const payload = validateVaultPayload(parseVaultPayload(await readPayload(source, options)));
+  const payload = validateVaultPayload(parseVaultPayload(await readPayload(source, options), source));
   await saveVaultEntry(definition, {
     tokens: payload.tokens,
     ...(payload.clientInfo ? { clientInfo: payload.clientInfo } : {}),
@@ -65,7 +67,7 @@ async function handleVaultClear(runtime: Pick<Runtime, 'getDefinition'>, args: s
   console.log(`Cleared OAuth vault entry for '${definition.name}'`);
 }
 
-function consumeVaultPayloadSource(args: string[]): { kind: 'file'; path: string } | { kind: 'stdin' } {
+function consumeVaultPayloadSource(args: string[]): VaultPayloadSource {
   const fileIndex = args.indexOf('--tokens-file');
   const stdinIndex = args.indexOf('--stdin');
   if (fileIndex !== -1 && stdinIndex !== -1) {
@@ -86,10 +88,7 @@ function consumeVaultPayloadSource(args: string[]): { kind: 'file'; path: string
   throw new CliUsageError('Usage: mcporter vault set <server> (--tokens-file <path> | --stdin)');
 }
 
-async function readPayload(
-  source: { kind: 'file'; path: string } | { kind: 'stdin' },
-  options: VaultCommandOptions
-): Promise<string> {
+async function readPayload(source: VaultPayloadSource, options: VaultCommandOptions): Promise<string> {
   if (source.kind === 'file') {
     return fs.readFile(source.path, 'utf8');
   }
@@ -107,12 +106,17 @@ async function readPayload(
   });
 }
 
-function parseVaultPayload(raw: string): unknown {
+function parseVaultPayload(raw: string, source: VaultPayloadSource): unknown {
   try {
     return JSON.parse(raw) as unknown;
-  } catch (error) {
-    const reason = error instanceof Error ? error.message : String(error);
-    throw new CliUsageError(`Vault payload is not valid JSON: ${reason}`);
+  } catch {
+    // The parser message quotes a prefix of the input, which for this command is a
+    // credential payload, so name the source instead of repeating the reason.
+    throw new CliUsageError(
+      source.kind === 'file'
+        ? `Vault payload file '${source.path}' is not valid JSON.`
+        : 'Vault payload from stdin is not valid JSON.'
+    );
   }
 }
 
@@ -152,11 +156,13 @@ function validateOAuthTokens(tokens: Record<string, unknown>): void {
       throw new CliUsageError(`Vault payload tokens.${key} must be a string.`);
     }
   }
-  if (
-    tokens.expires_in !== undefined &&
-    (!Number.isFinite(tokens.expires_in) || typeof tokens.expires_in !== 'number')
-  ) {
-    throw new CliUsageError('Vault payload tokens.expires_in must be a finite number.');
+  // expires_in, expires_at and expiresAt are what isStoredOAuthTokens reads back as
+  // finite numbers; a non-numeric one is dropped there together with the rest of the
+  // tokens, so reject it at the input boundary instead.
+  for (const key of ['expires_in', 'expires_at', 'expiresAt'] as const) {
+    if (tokens[key] !== undefined && (typeof tokens[key] !== 'number' || !Number.isFinite(tokens[key]))) {
+      throw new CliUsageError(`Vault payload tokens.${key} must be a finite number.`);
+    }
   }
 }
 
@@ -177,15 +183,22 @@ const CLIENT_INFO_STRING_ARRAY: ClientInfoFieldRule = {
 
 const CLIENT_INFO_FINITE_NUMBER: ClientInfoFieldRule = {
   description: 'a finite number',
-  matches: (value) => Number.isFinite(value),
+  matches: (value) => typeof value === 'number' && Number.isFinite(value),
 };
 
-// Field types follow OAuthClientInformationFullSchema (RFC 7591 dynamic client
-// registration) from @modelcontextprotocol/sdk/shared/auth.js, in its declaration
-// order. `client_id` is required by both members of OAuthClientInformationMixed and
-// is checked separately. Fields the schema leaves open (`jwks`) and provider
-// extensions outside it (`registration_client_uri`, `registration_access_token`) are
-// deliberately absent, so a complete registration response reaches the vault intact.
+// Field types follow the JSON types OAuthClientInformationFullSchema (RFC 7591
+// dynamic client registration) declares in @modelcontextprotocol/sdk/shared/auth.js,
+// in its declaration order, with `issuer` appended because mcporter stores it and
+// isStoredOAuthClientInformation reads it back as a string. URL-shaped fields are
+// only checked for being strings; the SDK's stricter SafeUrlSchema would reject
+// redirect URIs that are legal in the wild, such as urn:ietf:wg:oauth:2.0:oob.
+//
+// `client_id` is required by both members of OAuthClientInformationMixed and is
+// checked separately. `redirect_uris` stays optional even though the schema's wide
+// member requires it, because the narrow member, OAuthClientInformation, is equally
+// valid here. Fields the schema leaves open (`jwks`) and provider extensions outside
+// it (`registration_client_uri`, `registration_access_token`) are deliberately
+// absent, so a complete registration response reaches the vault intact.
 const CLIENT_INFO_FIELD_RULES: Readonly<Record<string, ClientInfoFieldRule>> = {
   redirect_uris: CLIENT_INFO_STRING_ARRAY,
   token_endpoint_auth_method: CLIENT_INFO_STRING,
@@ -205,6 +218,7 @@ const CLIENT_INFO_FIELD_RULES: Readonly<Record<string, ClientInfoFieldRule>> = {
   client_secret: CLIENT_INFO_STRING,
   client_id_issued_at: CLIENT_INFO_FINITE_NUMBER,
   client_secret_expires_at: CLIENT_INFO_FINITE_NUMBER,
+  issuer: CLIENT_INFO_STRING,
 };
 
 function validateOAuthClientInfo(clientInfo: Record<string, unknown>): void {
--- a/src/cli/vault-command.ts
+++ b/src/cli/vault-command.ts
@@ -235,6 +235,9 @@ export function printVaultHelp(): void {
     '',
     'Payload:',
     '  { "tokens": { "access_token": "...", "token_type": "Bearer" }, "clientInfo": { "client_id": "..." } }',
+    '',
+    '  clientInfo accepts a full dynamic client registration response, including',
+    '  redirect_uris, grant_types, response_types and the registration timestamps.',
   ];
   console.error(lines.join('\n'));
 }
--- a/src/cli/vault-command.ts
+++ b/src/cli/vault-command.ts
@@ -44,7 +44,7 @@ async function handleVaultSet(
     throw new CliUsageError(`Unknown vault set argument '${args[0]}'.`);
   }
   const definition = runtime.getDefinition(server);
-  const payload = validateVaultPayload(JSON.parse(await readPayload(source, options)));
+  const payload = validateVaultPayload(parseVaultPayload(await readPayload(source, options)));
   await saveVaultEntry(definition, {
     tokens: payload.tokens,
     ...(payload.clientInfo ? { clientInfo: payload.clientInfo } : {}),
@@ -105,6 +105,15 @@ async function readPayload(
     process.stdin.on('end', () => resolve(data));
     process.stdin.on('error', reject);
   });
+}
+
+function parseVaultPayload(raw: string): unknown {
+  try {
+    return JSON.parse(raw) as unknown;
+  } catch (error) {
+    const reason = error instanceof Error ? error.message : String(error);
+    throw new CliUsageError(`Vault payload is not valid JSON: ${reason}`);
+  }
 }
 
 function validateVaultPayload(value: unknown): VaultPayload {
--- a/src/cli/vault-command.ts
+++ b/src/cli/vault-command.ts
@@ -151,10 +151,66 @@ function validateOAuthTokens(tokens: Record<string, unknown>): void {
   }
 }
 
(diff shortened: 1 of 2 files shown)


Yigtwxx:
The one open item is the `client_id` requirement, so here is the measurement behind it.

Nothing in the repo seeds `clientInfo` without `client_id`. Searching `src/`, `tests/`, `docs/` and `README.md` finds no empty or `client_id`-less `clientInfo` object. The only internal writer, `saveClientInfo` (`src/oauth-persistence-stores.ts:384`), passes SDK registration output, which always carries one, and it does not go through this validator. `docs/config.md:259` documents the payload as `{ "tokens": { ... }, "clientInfo": { ... } }` without naming fields, so no documented workflow depends on the empty form.

So the tightening only reaches a hand-written script that seeds `clientInfo: {}`, and that script is already storing a truthy object which shadows the configured client and then fails during refresh with an opaque SDK error. If you would rather not take it, the drop is the three-line `client_id` block at the top of `validateOAuthClientInfo` plus the last row of the `rejects malformed clientInfo` table. Nothing else in the change depends on it.

steipete:
Superseded by the narrower compatibility-preserving fix in #288, now merged as [`4e8e37df2004aeef8b7c913b50d8de8dfecf1e32`](https://github.com/openclaw/mcporter/commit/4e8e37df2004aeef8b7c913b50d8de8dfecf1e32). That landing resolves #286 while retaining null tolerance, partial-object compatibility, unknown provider metadata, and existing acceptance of `clientInfo` without a mandatory `client_id`.

Thank you @Yigtwxx for the concurrent analysis and especially for identifying two separate worthwhile follow-ups. They are explicitly preserved outside this PR’s scope:

1. Sanitize malformed-JSON errors so credential payload fragments cannot be echoed in diagnostics.
2. Validate token `expires_at` and `expiresAt` values as finite numbers at the vault input boundary.

Closing this overlapping implementation as superseded; those follow-up findings are not being rejected or folded into the #286 fix.

PR #288, opened 2026-08-08
fix: accept OAuth DCR metadata in vault set

### LINKED ISSUE #286: vault set rejects clientInfo from OAuth dynamic client registration

feniix:
## Summary
- validate OAuth client information according to each known field's shape instead of requiring every value to be a string
- preserve standard DCR arrays and timestamps, unrestricted JWK/provider metadata, existing null-compatible imports, and the string-only workaround
- add command-level persistence and malformed-payload regression coverage

## Type of change
- [ ] Feature
- [x] Bug fix
- [ ] Hotfix
- [ ] Spike / exploration
- [ ] Documentation
- [ ] Refactor

## Real CLI behavior proof

Ran the built CLI with an isolated `MCPORTER_CONFIG` and `XDG_DATA_HOME`, the fake token from #286, DCR arrays and a registration timestamp. No server was contacted.

```text
$ printf '%s\n' "$PAYLOAD" | node dist/cli.js vault set demo --stdin
Saved OAuth credentials for 'demo' to <temp>/data/mcporter/credentials.json

$ jq '.entries[] | {tokens, clientInfo}
      | del(.tokens.__mcporter_generation, .clientInfo.__mcporter_client_generation)' \
    <temp>/data/mcporter/credentials.json
{
  "tokens": {
    "access_token": "fake",
    "token_type": "Bearer"
  },
  "clientInfo": {
    "client_id": "abc",
    "redirect_uris": [
      "https://example.test/cb"
    ],
    "grant_types": [
      "authorization_code"
    ],
    "response_types": [
      "code"
    ],
    "token_endpoint_auth_method": "none",
    "client_id_issued_at": 1754600000,
    "client_name": null
  }
}
```

This proves both the reported DCR shape and the previously accepted `null` value persist through the real CLI.

## Test plan
- `pnpm exec vitest run tests/vault-validation.test.ts tests/vault-command.test.ts`
- `pnpm check`
- `pnpm test`

Closes #286


feniix:
Addressed the review findings:

- restored null-compatible client-info imports with persistence regression coverage
- removed the release-owned changelog entry
- added a redacted real `vault set --stdin` persistence transcript to the PR body

@clawsweeper re-review

feniix:
It looks like #287 and this PR were developed concurrently and both address #286. Linking #287 here for visibility; happy for maintainers to choose or consolidate the preferred approach.

steipete:
Exact-head maintainer proof for `b91b1853cfcc5d6b6417c7b4ec6cecc8fd5f4f7a`:

- Rebased onto current `main` and retained Sebastian Otaegui / @feniix as author of all ten contributor commits. Added the maintainer-owned 0.13.1 changelog credit.
- Diff: 4 files, +106/-3 (`CHANGELOG.md`, `src/cli/vault-command.ts`, `tests/vault-command.test.ts`, `tests/vault-validation.test.ts`).
- Focused regression proof: `pnpm exec vitest run tests/vault-validation.test.ts tests/vault-command.test.ts` — 2 files, 26 tests passed.
- Repository gates: `pnpm docs:list`, `pnpm check`, `pnpm docs:site`, `pnpm test`, and `git diff --check` — all clean. Full suite: 183 files passed / 4 skipped; 1,389 tests passed / 26 skipped.
- Built CLI behavior proof: an isolated `node dist/cli.js vault set demo --stdin` run exited 0 and persisted fake RFC 7591 arrays, numeric issuance/secret-expiry timestamps, `client_name: null`, numeric token `expires_at`, and nested provider metadata with their original JSON values. A second fresh invocation with partial `clientInfo` and no `client_id` also exited 0. A malformed numeric `application_type` exited 1 with a field-specific error and did not replace the prior entry. The `example.test` server was not contacted.
- Structured Codex-backed autoreview: final integrated `autoreview --mode branch --base origin/main` run was clean with no accepted/actionable findings; secret scan clean. No review-driven changes were required.
- Exact diff/public-proof model-identifier audit: PASS; no model-bearing change or non-public identifier is present.
- Exact-head CI: [run 31277639023](https://github.com/openclaw/mcporter/actions/runs/31277639023) completed successfully for [Ubuntu](https://github.com/openclaw/mcporter/actions/runs/31277639023/job/93153735412), [macOS](https://github.com/openclaw/mcporter/actions/runs/31277639023/job/93153735393), and [Windows](https://github.com/openclaw/mcporter/actions/runs/31277639023/job/93153735417). PR state is mergeable/CLEAN.

Scope is intentionally limited to #286. Two useful findings from concurrent PR #287 remain separate follow-ups: sanitizing malformed-JSON errors and validating token `expires_at` / `expiresAt`. No live provider proof was needed because `vault set` only validates and persists the supplied local payload.

Thanks @feniix for the precise report, compatibility analysis, implementation, and real-CLI proof.

PR #306, opened 2026-08-10
fix(oauth): serialize the refresh transaction across processes

### LINKED ISSUE #305: OAuth refresh is not serialized across processes, allowing rotating-token replay and grant revocation
## Summary

mcporter serializes credential-file writes, but it does not serialize the complete OAuth refresh transaction:

1. Read cached tokens.
2. Redeem the refresh token.
3. Persist the rotated tokens.

When multiple mcporter processes share a credential vault, they can read the same expired token and concurrently redeem the same rotating refresh token.

Against providers such as Notion MCP, this produces `invalid_grant` responses and can trigger refresh-token replay detection, revoking the active token family. The user is then left with:

> Server '<name>' uses refreshable_bearer auth but has no cached access token.

This reproduces in mcporter 0.13.3 and current `main` (`7f48e5d`).

## Root cause

`readExplicitRefreshableBearerToken` performs the token request and persistence as separate, unlocked operations:

https://github.com/openclaw/mcporter/blob/v0.13.3/src/oauth-token-refresh.ts#L257-L305

```ts
const refreshed = await refreshBearerToken(definition, tokens.refresh_token);
await persistence.saveTokens(refreshed);
```

`saveVaultEntry` locks the vault only while writing:

https://github.com/openclaw/mcporter/blob/v0.13.3/src/oauth-vault.ts#L307-L329

This prevents file corruption but does not prevent two processes from redeeming the same refresh-token generation.

The rejected-refresh recovery added for #226 compare-clears stale local credentials and can adopt a newer locally persisted generation:

https://github.com/openclaw/mcporter/blob/v0.13.3/src/oauth-token-refresh.ts#L215-L254

That recovery is useful defense in depth, but it runs after the refresh-token replay has reached the provider. It cannot undo provider-side revocation.

## Deterministic reproduction

I tested the installed 0.13.3 build using:

- One expired cached access token.
- One valid refresh token.
- Twelve fresh Node processes sharing the same persistence directory.
- A mock OAuth provider that rotates the refresh token on first use and treats reuse as a replay, revoking the active family.
- The default refresh skew; newly issued tokens were not left inside the refresh window.

Across three runs without a refresh lock:

```text
run 1: 12 callers, 4 token requests, 3 replays
run 2: 12 callers, 6 token requests, 5 replays
run 3: 12 callers, 12 token requests, 11 replays
```

Every run produced:

```text
family revoked: true
next sequential refresh: invalid_grant
final access token present: false
final refresh token present: false
```

Some losing processes adopted the locally persisted winner, but the provider had already revoked that winner's token family because of the replay.

## Locking experiment

I repeated the same test while wrapping the complete transaction in a separate cross-process file lock:

```text
acquire credential refresh lock
re-read persisted tokens
check whether refresh is still required
refresh using the latest generation
persist access + refresh tokens
release lock
```

Results across three runs:

```text
12/12 callers succeeded
1 refresh request during the concurrent wave
0 replay attempts
token family remained valid
a later sequential refresh succeeded
access and refresh tokens remained persisted
```

This validates transaction-level cross-process locking as the fix.

## Provider behavior

Notion documents that:

- Access tokens expire after one hour.
- Refresh tokens rotate.
- Concurrent refreshes can cause later requests to receive `invalid_grant`.
- `invalid_grant` must not be retried.

https://github.com/makenotion/notion-cookbook/blob/main/docs/mcp-client-integration.md#token-refresh-fails-with-invalid_grant

RFC 9700 §4.14.2 describes refresh-token rotation and notes that replay detection may revoke the active refresh token:

https://www.rfc-editor.org/rfc/rfc9700.html#section-4.14.2

## Existing test gap

The current cross-process refresh integration test starts only one refreshing process, and its mock endpoint accepts every refresh request:

https://github.com/openclaw/mcporter/blob/v0.13.3/tests/oauth-refresh-process.integration.test.ts#L50-L58

It verifies that DCR metadata survives a refresh, but cannot detect concurrent redemption of a single-use refresh token.

## Suggested fix

Add a dedicated cross-process refresh lock per credential identity.

The lock must be separate from the vault write lock; using the vault lock would deadlock when the refresh path calls `saveTokens`.

Under the refresh lock:

1. Re-read tokens from persistence.
2. If the generation changed and the latest access token is outside the refresh window, return it.
3. Otherwise redeem the latest persisted refresh token.
4. Persist the returned access and refresh tokens before releasing the lock.
5. Retain the existing generation-aware `invalid_grant` recovery as defense in depth.

Apply the same protection to both refresh paths:

- `auth: "oauth"`
- `auth: "refreshable_bearer"`

The lock key should identify the credential, not the refresh-token value, and should not expose secrets in filenames or logs.

## Regression coverage

Add a built-artifact multi-process test that:

1. Seeds one expired token generation.
2. Starts multiple fresh processes simultaneously.
3. Uses a mock provider that accepts the first refresh-token redemption, rotates it, and revokes the family on replay.
4. Asserts that the provider receives exactly one request for the original generation.
5. Asserts that every caller receives the winner's access token.
6. Asserts that the rotated refresh token remains persisted.
7. Forces a later legitimate refresh and verifies it succeeds.
8. Covers stale-lock recovery when a refreshing process exits unexpectedly.

A second case should verify that unrelated credential identities can refresh concurrently.

## Related issues

- #226 added `invalid_grant` classification, concurrent-winner adoption, and compare-and-clear recovery. Its optional refresh serialization hardening remains unimplemented.
- #167 serialized credential-file writes, but explicitly did not serialize the OAuth protocol transaction.

feniix:
## Summary

Fixes #305. Concurrent mcporter processes sharing a credential vault could each read the same expired access token and redeem the same rotating refresh token. Providers with replay detection answer `invalid_grant` and revoke the whole token family, leaving no cached credentials.

A per-credential cross-process lock now covers the whole refresh transaction — re-read, redeem, persist — so a rotating token is redeemed once.

## What changed

**The lock.** New `src/oauth-refresh-lock.ts`, built on the existing `withFileLock` (no new dependency), on a target separate from the vault write lock so the locked section can still persist. It is re-entrant per call chain, because the MCP SDK reads tokens through the provider while an outer wrapper holds the lock.

**One credential, one lock.** A credential reachable from more than one place locks every one of them: the token-cache store alongside the vault, and the legacy `<name>-oauth` vault entry a renamed server inherits from. Two configurations need not derive identical lock sets — they must overlap on the target holding the credential they share — and the set is sorted so overlapping sets cannot invert. Identities are canonicalized and hashed, and each filename derives from its own target rather than the calling definition; otherwise two configurations sharing one key hash it into two filenames and take separate locks anyway.

**All three redemption paths** hold it: the cached-read path, the `refreshable_bearer` path, and SDK-driven refresh via the provider's token read.

**Divergent stores.** The composite read returns the first store holding tokens, so a partly-landed save can leave a spent generation shadowing a newer one. The transaction now redeems the newest. Two limits are load-bearing: it overrides store priority *only* when the stores hold different refresh tokens, and it never writes the winner back — flattening the stores to one generation would make rejected-credential recovery clear every copy and discard a generation that should survive.

**Nothing redeemable escapes the lock.** On lock timeout the cached-read path returns its persisted access token, which a caller cannot redeem. The provider path cannot do the same, because the token *object* it returns is redeemable by the SDK, so it fails with a retryable error instead. Withholding just the refresh token is not an option: the SDK re-persists the provider's return value when `issuer` is unset and would destroy it.

**Bounded requests.** Token-endpoint and discovery calls made under the lock carry a cancellation deadline. A live-but-stalled holder is invisible to pid-based stale detection, so an unbounded call would starve every waiter.

**Rejected-grant classification.** The SDK reports OAuth error codes on `code`, which the classifier did not read — so a genuinely rejected grant was treated as transient and never reached the recovery that clears dead credentials.

## Proof

**Deterministic.** A mock provider that rotates on redemption and revokes the family on replay. Twelve concurrent processes against one seeded expired generation, on both the cached-read and provider paths: exactly one redemption, zero replays, family intact, and a later sequential refresh still succeeding. Run three times per path. Unpatched 0.13.3 produced 4/6/12 redemptions with 3/5/11 replays across the same waves and revoked the family every time.

Also covered: unrelated identities refreshing concurrently (proven by holding both token requests open until each has arrived, not by ordering), a held lock, a lock left by a dead process, and a crash between redeeming and persisting.

**Live.** A disposable Notion account on a real host, using this PR's build at `8fdb6d8`. Twelve concurrent processes: 12/12 succeeded, no `invalid_grant`, generation rotated A → B with both tokens still present. Generation B was then forced due and refreshed to C — Notion accepted the refresh token the concurrent wave had persisted, which is what separates a wave that worked from one that silently left a revoked family. Generations are labelled A/B/C; the capture contains no token, authorization header, client identifier, secret, workspace identifier, or endpoint.

Boundaries on the live trace: both rotations were forced via the refresh skew rather than waiting out a natural token lifetime; it carries no provider-side request logs, so redemption counts come from the deterministic test rather than this capture; and it predates the divergent-store and lock-identity commits, whose code paths that single-store scenario does not exercise.

<details>
<summary>Redacted live capture (full trace, commit <code>8fdb6d8</code>)</summary>

Captured: 2026-08-10T18:06:15Z · Tested commit: `8fdb6d81f0369629a169e3c302cfb848df7396fb`

**Environment**

- Provisioned host: Darwin 25.5.0 arm64
- Node.js: `v24.18.0`, pnpm: `10.34.5`
- Account: disposable Notion OAuth account `notion:refresh-test`

The PR was built locally and its `dist/cli.js` was run against the provisioned host's real mcporter credential vault. A temporary copy of the mcporter config increased `refresh.refreshSkewSeconds` for this account so refresh was due immediately. The provisioned config itself was not modified.

**Redaction policy:** no access token, refresh token, authorization header, OAuth client identifier, client secret, tenant/workspace identifier, endpoint URL, or raw tool response appears below. Credential generations are represented as A, B, and C. Process output is reduced to exit code, status counts, and error classification.

**Preflight** — at `2026-08-10T18:02:32.732Z` the live vault held one credential entry:

```json
{
  "generation": "A",
  "expiresAt": "2026-08-11T02:02:32Z",
  "hasAccessToken": true,
  "hasRefreshToken": true,
  "hasFullDynamicClientMetadata": true
}
```

A sequential status probe before forcing expiry returned `status: "ok"`.

**Concurrent live refresh wave** — twelve fresh PR-built processes started concurrently, each equivalent to:

```sh
HOME=<agent-home> \
MCPORTER_CONFIG=<temporary-forced-refresh-config> \
<pr-306-build>/dist/cli.js \
  list notion:refresh-test --status --json --no-oauth
```

```json
{
  "callers": 12,
  "successfulCallers": 12,
  "failedCallers": 0,
  "invalidGrantObservations": 0,
  "statuses": [{ "status": "ok", "count": 12 }],
  "perCaller": ["1:ok","2:ok","3:ok","4:ok","5:ok","6:ok","7:ok","8:ok","9:ok","10:ok","11:ok","12:ok"]
}
```

After the wave, the live vault showed:

```json
{
  "updatedAt": "2026-08-10T18:05:36.911Z",
  "generationTransition": "A -> B",
  "expiresAt": "2026-08-11T02:05:36Z",
  "expiryAdvanced": true,
  "hasAccessToken": true,
  "hasRefreshToken": true
}
```

No process emitted stderr. No output contained `invalid_grant`, `no cached access token`, or an unauthenticated status.

**Subsequent rotated-generation refresh** — the temporary config was adjusted so generation B was immediately due, then one fresh process ran the same status probe:

```json
{
  "exitCode": 0,
  "okCount": 1,
  "authCount": 0,
  "errorCount": 0,
  "serverStatus": "ok",
  "invalidGrant": false
}
```

```json
{
  "updatedAt": "2026-08-10T18:06:03.380Z",
  "generationTransition": "B -> C",
  "expiresAt": "2026-08-11T02:06:03Z",
  "expiryAdvanced": true,
  "hasAccessToken": true,
  "hasRefreshToken": true
}
```

This second rotation is the grant-health check: Notion accepted the refresh token persisted by the concurrent wave and returned another usable generation, so the wave did not leave a replay-revoked family or purge the cached credentials.

Full capture also posted as a [PR comment](https://github.com/openclaw/mcporter/pull/306#issuecomment-5246819744).

</details>

## Known limitation

A 401 makes the SDK redeem whatever the provider's token read returns, even a token that was just refreshed, and that call site has no seam to wrap. Two processes taking a 401 at the same instant can still redeem concurrently; the rejected-refresh recovery remains the backstop. This is documented in the provider code. The missing expiry check is worth reporting upstream, since it also burns a token rotation on every `auth()` call.

Two operational notes for adopters: the guarantee is per host, because pid-based stale detection cannot arbitrate between hosts sharing one credential directory over a network filesystem; and a mixed-version fleet is only fully protected once every long-lived daemon and generated CLI is upgraded.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [0.13.4] - Unreleased
 
+### OAuth
+
+- Serialize the whole OAuth refresh transaction across processes so a rotating refresh token is redeemed once. Concurrent mcporter processes sharing a credential vault could each read the same expired access token and redeem the same refresh token; providers that detect replay answer with `invalid_grant` and revoke the entire token family, leaving no cached credentials. A dedicated per-credential refresh lock now covers the re-read, redemption, and persist as one transaction, and a caller that finds another process already refreshed adopts that result instead of redeeming again. Token endpoint and discovery requests made under the lock carry a cancellation deadline so a stalled endpoint cannot starve other processes. The guarantee is per host: hosts sharing one credential directory over a network filesystem cannot arbitrate each other's locks, and every long-lived daemon or generated CLI has to be upgraded before a fleet is fully protected. One known gap remains by design — the MCP SDK's internal 401-driven refresh has no interception point, so a 401 reaching two processes at the same instant can still redeem concurrently, where the existing rejected-refresh recovery remains the backstop. (Issue #305, thanks @feniix)
+- Recognize the OAuth error code the MCP SDK reports on its error objects, so a genuinely rejected refresh grant reaches the recovery that clears dead credentials instead of being retried as a transient failure. (Issue #305, thanks @feniix)
+
 ## [0.13.3] - 2026-08-09
 
 ### CLI
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 ## [0.13.4] - Unreleased
 
-### OAuth
-
-- Serialize the whole OAuth refresh transaction across processes so a rotating refresh token is redeemed once. Concurrent mcporter processes sharing a credential vault could each read the same expired access token and redeem the same refresh token; providers that detect replay answer with `invalid_grant` and revoke the entire token family, leaving no cached credentials. A dedicated per-credential refresh lock now covers the re-read, redemption, and persist as one transaction, and a caller that finds another process already refreshed adopts that result instead of redeeming again. Token endpoint and discovery requests made under the lock carry a cancellation deadline so a stalled endpoint cannot starve other processes. The guarantee is per host: hosts sharing one credential directory over a network filesystem cannot arbitrate each other's locks, and every long-lived daemon or generated CLI has to be upgraded before a fleet is fully protected. One known gap remains by design — the MCP SDK's internal 401-driven refresh has no interception point, so a 401 reaching two processes at the same instant can still redeem concurrently, where the existing rejected-refresh recovery remains the backstop. (Issue #305, thanks @feniix)
-- Recognize the OAuth error code the MCP SDK reports on its error objects, so a genuinely rejected refresh grant reaches the recovery that clears dead credentials instead of being retried as a transient failure. (Issue #305, thanks @feniix)
-
 ## [0.13.3] - 2026-08-09
 
 ### CLI
--- a/src/fs-json.ts
+++ b/src/fs-json.ts
@@ -11,6 +11,23 @@ const MAX_SYMLINK_DEPTH = 40;
 const DEFAULT_ATOMIC_FILE_MODE = 0o600;
 const localLockTails = new Map<string, Promise<void>>();
 
+// Callers that degrade on contention (the OAuth refresh lock returns a stale
+// token instead of redeeming unlocked) must tell a timeout apart from a real
+// filesystem failure. Message matching would misroute either direction.
+export class FileLockTimeoutError extends Error {
+  readonly lockPath: string;
+
+  constructor(lockPath: string, options?: ErrorOptions) {
+    super(`Timed out waiting for file lock ${lockPath}`, options);
+    this.name = 'FileLockTimeoutError';
+    this.lockPath = lockPath;
+  }
+}
+
+export function isFileLockTimeoutError(error: unknown): error is FileLockTimeoutError {
+  return error instanceof FileLockTimeoutError;
+}
+
 // readJsonFile reads a JSON file and returns undefined when the file does not exist.
 export async function readJsonFile<T = unknown>(filePath: string): Promise<T | undefined> {
   try {
@@ -95,7 +112,7 @@ export async function withFileLock<T>(
           continue;
         }
         if (Date.now() - startedAt > timeoutMs) {
-          throw new Error(`Timed out waiting for file lock ${lockPath}`, { cause: error });
+          throw new FileLockTimeoutError(lockPath, { cause: error });
         }
         await sleep(LOCK_POLL_MS);
       }
@@ -149,10 +166,7 @@ async function waitForLocalLock(previous: Promise<void>, timeoutMs: number, key:
     await Promise.race([
       previous,
       new Promise<never>((_, reject) => {
-        timer = setTimeout(
-          () => reject(new Error(`Timed out waiting for file lock ${key}.lock`)),
-          Math.max(0, timeoutMs)
-        );
+        timer = setTimeout(() => reject(new FileLockTimeoutError(`${key}.lock`)), Math.max(0, timeoutMs));
       }),
     ]);
   } finally {
@@ -176,6 +190,12 @@ async function resolveAtomicWriteTarget(filePath: string): Promise<{ path: strin
     }
     throw error;
   }
+}
+
+// Resolves a path to its real location so two spellings of one file (relative,
+// absolute, or symlinked) cannot be treated as separate identities.
+export async function canonicalizeFilePath(filePath: string): Promise<string> {
+  return await resolvePathFollowingSymlinks(path.resolve(filePath));
 }
 
 async function resolvePathFollowingSymlinks(filePath: string): Promise<string> {
--- a/src/oauth-persistence-stores.ts
+++ b/src/oauth-persistence-stores.ts
@@ -507,6 +507,11 @@ export class CompositePersistence implements OAuthPersistence {
     return result.value;
   }
 
+  async readTokensPerStore(): Promise<OAuthTokens[]> {
+    const perStore = await Promise.all(this.stores.map((store) => store.readTokens()));
+    return perStore.filter((tokens): tokens is OAuthTokens => tokens !== undefined);
+  }
+
   async saveTokens(tokens: OAuthTokens): Promise<void> {
     // Compute the absolute expiry once so every backing store records the same
     // generation value even at a wall-clock second boundary.
--- a/src/oauth-persistence.ts
+++ b/src/oauth-persistence.ts
@@ -27,6 +27,12 @@ export interface OAuthPersistence {
   describe(): string;
   readSnapshot(): Promise<OAuthPersistenceSnapshot>;
   readTokens(): Promise<StoredOAuthTokens | undefined>;
+  // Every backing store's own view of the tokens. readTokens() returns the
+  // first store that has any, which hides divergence: a partial save can leave
+  // a spent generation in the higher-priority store while a newer one sits
+  // behind it. Refresh reconciles across these before redeeming. Only the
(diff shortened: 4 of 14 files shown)


feniix:
@clawsweeper re-review

Head is now `031ca97`. Since the last review both remaining findings are fixed and the proof ask is addressed in the PR body:

- **Lock inherited legacy OAuth credentials under one identity (P1)** — the lock set now includes every vault key that can supply a definition's credentials, so a renamed server and the legacy `<name>-oauth` entry it inherits overlap on the shared key. Sets are sorted to keep overlapping sets from inverting, and each lock filename derives from its own target rather than the calling definition. Concurrent rename-path regression test added, verified failing without the fix.
- **Real behavior proof** — a live run against a disposable Notion account is summarized under Proof: twelve concurrent processes all succeeded with no `invalid_grant`, the generation rotated, and a subsequent forced refresh of the rotated generation succeeded, which is the grant-health check. Findings are described in prose rather than attached because the capture lives in a non-public store; its boundaries are stated in the body.

Also since the prior review: divergent vault/cache generations are reconciled under the lock (selection only, never write-back, and only when the stores hold different refresh tokens), and the release-owned `CHANGELOG.md` edit is gone.

CI is green on Ubuntu, macOS, and Windows.

feniix:
@clawsweeper re-review

steipete:
Maintainer follow-up `1a6bca8` is pushed to this branch.

It fixes the remaining missing-client regression: an expired orphaned token set is cleared when the SDK observes that no dynamic client registration exists, allowing normal DCR and interactive authorization instead of raising refresh contention. It also strengthens process coverage for a real holder killed by `SIGKILL` and for a transient refresh failure followed by a successful fresh process.

Local proof:

```text
$ pnpm check
All matched files use the correct format.
Found 0 warnings and 0 errors.
typecheck: passed

$ pnpm test
Test Files  190 passed | 4 skipped (194)
Tests       1487 passed | 26 skipped (1513)

$ node dist/cli.js --version
0.13.3

$ node /tmp/mcporter-oauth-live-proof.mjs "$PWD"
race proof: callers=12 requests=1 replays=0 revoked=false access_generations=1
reauth proof: sdk_result=REDIRECT registrations=1 prompts=1 orphaned_tokens_remaining=false
```

The last command was a one-off standalone built-artifact integration harness using the committed process fixture and a rotating mock OAuth provider; it was removed after the run.

Structured Codex review was clean for both the follow-up delta and the complete branch against `origin/main`.

CI for head `1a6bca8e5aa8228f6a6c2ef7afccb57f3ba69248` is green on Ubuntu, macOS 15, and Windows: https://github.com/openclaw/mcporter/actions/runs/31449917875

Verdict: ready to land after maintainer approval. I did not merge it.

PR #308, opened 2026-08-12
fix(chrome): discover OpenClaw relay endpoint

steipete:
## Summary

- discover OpenClaw's active Chrome extension relay endpoint with `openclaw browser extension cdp --json`
- preserve precedence: explicit `MCPORTER_CHROME_DEVTOOLS_RELAY_URL`, strict safe metadata discovery, then the `127.0.0.1:18799` compatibility fallback
- accept only the exact loopback Browser Relay Authentication v2 contract with a matching local `keyId`; never consume legacy bearer headers
- mirror OpenClaw named-profile credential paths and environment precedence
- include the asynchronously resolved endpoint in versioned daemon runtime identity so profile, key, policy, and relay-port changes retire stale daemon state
- resolve Windows `openclaw.cmd` only from safe absolute PATH entries and terminate the full process tree on bounded timeout or output overflow

## Proof

- focused relay/client/proxy/handoff/daemon gate: 279 passed, 1 skipped
- full suite on current `origin/main`: 1,594 passed, 26 skipped
- `pnpm check` and `git diff --check` clean
- final max-P1 autoreview and TruffleHog clean
- live OpenClaw proof: Gateway on `19100`, connected extension relay on `19110`, no listener on `18799`; built mcporter with the explicit URL override removed succeeded under `policy=require` and found a public `example.com` tab, proving metadata discovery selected `19110` rather than the compatibility fallback

No release, publish, tag, or deployment work is included.


steipete:
@clawsweeper re-review

Exact-head CI is green on `d1a3def`, including Windows. The changelog entry is intentionally retained: this repository's maintainer instructions require a changelog entry for user-facing fixes, and the landing request explicitly requires preserving it.

PR #282, opened 2026-08-06
fix(transport): avoid stall when legacy SSE headers stay pending

Quack6765:
## What Problem This Solves

MCPorter 0.13.0 can finish OAuth successfully and still report a legacy HTTP MCP server as offline. After `notifications/initialized`, the SDK v2 client starts a long-lived standalone SSE `GET`; MCPorter waits for that fetch promise before allowing `list` or `call` to continue. A fetch promise does not settle until response headers arrive, so a server that accepts the SSE request but leaves its response completely idle can hold MCPorter at startup until the outer command timeout.

That is the live Mobbin failure behind this PR: authentication completes, but `mcporter list mobbin` times out after 30 seconds. This is distinct from #234, which isolates a byte-idle SSE response that has already sent headers from the ordinary request connection pool. The unbounded pre-header wait was added with the SDK v2 migration in #255.

## Why This Change Was Made

The repair is deliberately narrow. Legacy connections still give the standalone SSE receive channel a chance to establish, but that readiness wait is capped at 250 ms:

- if response headers arrive sooner, MCPorter continues immediately;
- if headers remain pending, MCPorter continues after 250 ms;
- the SSE request is not cancelled and remains open in the background for later server-to-client messages.

Removing the readiness wait entirely made the existing legacy HTTP elicitation/sampling fixture race its first tool call ahead of the receive channel. Keeping the wait unbounded reproduces the reported stall. The bounded grace passes both paths without adding configuration, dependencies, or changes to OAuth behavior.

The pending-header regression fixture accepts the standalone SSE request without flushing response headers, matching the observed hosted-server behavior. A second compatibility test routes a real legacy fixture through a proxy that forwards the SSE request immediately but withholds its response headers for 500 ms—twice the startup grace. Its first tool call begins after the grace, receives a later `elicitation/create` request on the still-open stream, returns the headless decline, and completes successfully. This directly covers the delayed-header ordering boundary raised by ClawSweeper.

The production change remains confined to `src/runtime/http-transport.ts`; the other changes are focused integration coverage and the 0.13.1 changelog entry.

## User Impact

Legacy hosted MCP servers such as Mobbin can list and call tools even when their standalone SSE response does not immediately send headers. Servers that establish the receive channel normally retain the existing readiness ordering. When headers arrive after the grace, the receive stream remains active and can still deliver server-initiated elicitation, as verified with a 500 ms delay. Header-idle servers incur at most the 250 ms startup grace per new connection instead of hanging until the command timeout.

[...]

PR #311, opened 2026-08-12
fix(http): stop SSE fallback from masking primary transport errors

### LINKED ISSUE #310: SSE fallback fires on any primary error, replacing the real cause with a misleading HTTP 405 on streamable-HTTP-only servers
## Summary

When the primary StreamableHTTP connection fails for a reason that has nothing to do with transport negotiation — a transient `fetch failed`, a connect timeout, a DNS stall — `attemptHttpClientContext` still falls back to the legacy SSE transport. Against a server that only speaks streamable HTTP, that fallback attempt does a GET on a POST-only endpoint and gets a `405`. The `405` is then what surfaces to the user, and the actual cause is gone.

The result is a confident, specific, and completely wrong error message. It reads as "the server rejected the method", so you go debug the server and the endpoint URL, when the real problem was local networking.

## Environment

- mcporter `0.13.4` (also reproduced on `0.8.1`)
- Node 24.x, Linux (WSL2)
- Server: `https://mcp.exa.ai/mcp` — streamable-HTTP only, no auth, no API key

Config is minimal:

```json
{ "mcpServers": { "exa": { "baseUrl": "https://mcp.exa.ai/mcp" } } }
```

## What we saw

```
$ mcporter call "exa.web_search_exa(query: \"test\", numResults: 1)"
[mcporter] exa.web_search_exa responded with HTTP 405 (SSE error: Non-200 status code (405)).
[mcporter] exa.web_search_exa responded with HTTP 405 (SSE error: Non-200 status code (405)).
[mcporter] SSE error: Non-200 status code (405)
SseError: SSE error: Non-200 status code (405)
    at _eventSource.onerror (.../@modelcontextprotocol/sdk/dist/esm/client/sse.js:93:31)
```

Reproducible on every call. What made it especially misleading: **`mcporter list exa` succeeded at the same time**, printing the schema for both tools in ~2.5s. A server that lists fine but 405s on every call points you squarely at the call path or the endpoint, which is the wrong place entirely.

The actual cause turned out to be local: DNS on this box was taking ~10.0s against a 10s connect timeout, so the primary transport intermittently lost a coin flip. Nothing about the error said so.

## Why it happens

In `attemptHttpClientContext`, the `catch (primaryError)` block has three escape hatches — aborted signal, `isEraNegotiationFailure`, and `shouldAbortSseFallback` — and then falls back to SSE **unconditionally**:

```js
catch (primaryError) {
    if (options.signal?.aborted || isEraNegotiationFailure(primaryError)) { ...throw primaryError; }
    if (shouldAbortSseFallback(primaryError)) { ...throw primaryError; }
    if (isUnauthorizedError(primaryError)) { ... }
    if (primaryError instanceof Error) {
        logger.info(`Falling back to SSE transport for '${activeDefinition.name}': ${primaryError.message}`);
    }
    return { context: await connectSseFallbackTransport(...) };
}
```

And `shouldAbortSseFallback` only consults the transport-mismatch predicate on the post-auth path:

```js
function shouldAbortSseFallback(error) {
    if (isPostAuthConnectError(error))
        return !isLegacySseTransportMismatch(error);
    return isOAuthFlowError(error) || error instanceof OAuthTimeoutError;
}
```

So for a plain network error that is neither an OAuth error nor a post-auth connect error, `shouldAbortSseFallback` returns `false` and the SSE attempt proceeds.

The interesting part is that the codebase already has exactly the right predicate. `isLegacySseTransportMismatch` tests for `404`/`405`, which is the genuine "this server is the old SSE kind" signal — it just is not applied on the ordinary path. `isEraNegotiationFailure` is already treated as not-worth-a-fallback for essentially this reason; the same logic seems to extend to any error that is not a transport mismatch.

There is a second, separable problem: even when the fallback is legitimate, the SSE attempt's error **replaces** the primary error rather than being attached to it. That is what turns a diagnosable fault into a misleading one.

## Suggested fix

Two options, and the first is safe regardless of what you decide about the second:

1. **Stop the fallback error from masking the primary error.** When the SSE fallback also fails, report the primary error as the cause (with the fallback failure as context, e.g. `cause`). Even keeping today's fallback behavior exactly as-is, this alone would have made the real problem visible immediately.

2. **Only fall back when the primary error actually looks like a legacy-SSE mismatch** — i.e. gate the fallback on `isLegacySseTransportMismatch(primaryError)` on the ordinary path too, not just the post-auth one. Optionally retry the primary transport a bounded number of times for transient errors before giving up.

I have been running option 2 locally as a patch and it has been stable (10/10 successful calls immediately after, versus intermittent false-405s before). But option 2 is a real behavior change and I may well be missing servers whose legacy-SSE mismatch presents as something other than 404/405 — that judgment is yours, which is why I am filing this rather than opening a PR blind.

Happy to send a PR for either shape if you have a preference.

## Repro

Any streamable-HTTP-only MCP endpoint plus an unreliable primary connection will do. To force it deterministically without waiting for a flaky network, make the first StreamableHTTP attempt throw a generic error (e.g. a non-HTTP `fetch failed`) and observe that the reported error becomes the SSE `405` rather than the injected cause.

SebTardif:
## What Problem This Solves

When Streamable HTTP connect fails for a generic network reason (`fetch failed`, DNS stall, connect timeout), mcporter still fell back to legacy SSE. Against streamable-HTTP-only servers, that fallback GET commonly returns 405 and replaced the real primary error with a misleading SSE failure.

## Summary

Closes #310.

- Ordinary Streamable HTTP errors now surface unchanged and do not trigger legacy SSE.
- Explicit 404/405 transport mismatches still select the legacy SSE transport.
- Existing 401/OAuth promotion and fallback paths remain intact.
- Once a confirmed mismatch selects SSE, a later SSE failure remains the actionable error.

## Real behavior proof

Built from this PR head and exercised against real local HTTP endpoints through the public runtime API:

1. A server that resets the Streamable HTTP POST and would return 405 to a fallback GET:
   - `origin/main`: request trace was `POST /mcp`, then `GET /mcp`; surfaced `SSE error: Non-200 status code (405)`.
   - this PR: request trace was only `POST /mcp`; surfaced the original `TypeError: fetch failed` with native cause `other side closed`.
   - repeated twice with the same POST-only result.
2. A legacy endpoint that returns 405 to Streamable HTTP and 503 to SSE:
   - request trace was `POST /mcp`, then `GET /mcp`;
   - surfaced `SSE error: Non-200 status code (503)`, proving legitimate mismatch fallback remains active and its operational failure is preserved.

## Regression proof

The pending-header and delayed-header proofs from #282 remain green:

```console
$ pnpm exec vitest run tests/cli-idle-sse.integration.test.ts --reporter=verbose
Test Files  1 passed (1)
Tests       2 passed (2)

$ pnpm exec vitest run tests/e2e-fixture-servers.test.ts --reporter=verbose --testTimeout=30000 -t 'headers arrive after the startup grace'
Test Files  1 passed (1)
Tests       1 passed | 14 skipped (15)
```

Focused transport and composability coverage:

```console
$ pnpm exec vitest run tests/runtime-transport.test.ts tests/runtime-compose.test.ts --reporter=verbose
Test Files  2 passed (2)
Tests       63 passed (63)
```

Full repository gates:

```console
$ pnpm check
# format, OXLint, and TypeScript checks passed

$ pnpm test
Test Files  191 passed | 4 skipped (195)
Tests       1598 passed | 26 skipped (1624)
```

### DIFF
--- a/src/runtime/http-transport.ts
+++ b/src/runtime/http-transport.ts
@@ -156,21 +156,6 @@ function shouldAbortSseFallback(error: unknown): boolean {
   return !isLegacySseTransportMismatch(error);
 }
 
-/**
- * Prefer the primary Streamable HTTP error when SSE fallback also fails.
- * Attaches the SSE failure as `cause` so diagnostics stay available without
- * replacing a network timeout with a misleading legacy-SSE 405.
- */
-function preferPrimaryTransportError(primaryError: unknown, sseError: unknown): unknown {
-  if (primaryError instanceof Error) {
-    if (sseError instanceof Error && primaryError.cause === undefined) {
-      primaryError.cause = sseError;
-    }
-    return primaryError;
-  }
-  return sseError;
-}
-
 function isEraNegotiationFailure(error: unknown): boolean {
   return !!error && typeof error === 'object' && 'code' in error && error.code === SdkErrorCode.EraNegotiationFailed;
 }
@@ -283,8 +268,7 @@ async function attemptHttpClientContext(
         logger,
         options,
         wrapRecordTransport,
-        clientFactory,
-        primaryError
+        clientFactory
       ),
     };
   }
@@ -328,8 +312,7 @@ async function connectSseFallbackTransport(
   logger: Logger,
   options: CreateClientContextOptions,
   wrapRecordTransport: WrapRecordTransport,
-  clientFactory: HttpClientFactory,
-  primaryError?: unknown
+  clientFactory: HttpClientFactory
 ): Promise<ClientContext> {
   const createSseTransport = () =>
     wrapRecordTransport(new SSEClientTransport(command.url, transportOptions), definition, options);
@@ -355,16 +338,10 @@ async function connectSseFallbackTransport(
       // Auth challenges from SSE (including ad-hoc HTTP discovery) stay authoritative.
       throw sseError;
     }
-    // Prefer primary so a failed SSE GET (often 405 on streamable-only) never erases
-    // the Streamable HTTP fault that triggered the fallback (#310).
-    if (primaryError !== undefined) throw preferPrimaryTransportError(primaryError, sseError);
     throw sseError;
   }
 }
 
 export const __test = {
   waitForStandaloneSseStart,
-  isLegacySseTransportMismatch,
-  shouldAbortSseFallback,
-  preferPrimaryTransportError,
 };
--- a/src/runtime/http-transport.ts
+++ b/src/runtime/http-transport.ts
@@ -144,8 +144,31 @@ async function closeOAuthSession(oauthSession?: OAuthSession): Promise<void> {
 }
 
 function shouldAbortSseFallback(error: unknown): boolean {
+  // After a completed Streamable auth challenge: only fall back on 404/405
+  // transport-mismatch signals (legacy SSE servers), not on other post-auth faults.
   if (isPostAuthConnectError(error)) return !isLegacySseTransportMismatch(error);
-  return isOAuthFlowError(error) || error instanceof OAuthTimeoutError;
+  if (isOAuthFlowError(error) || error instanceof OAuthTimeoutError) return true;
+  // 401/auth still needs the promote + SSE paths below; do not short-circuit them.
+  if (isUnauthorizedError(error)) return false;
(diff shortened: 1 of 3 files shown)


SebTardif:
@clawsweeper re-review

PR #325, opened 2026-08-21
fix(cli): use canonical URL serialization

vincentkoc:
## Summary

- return canonical `URL.href` directly
- cover origin, trailing-slash path, and non-trailing-slash path forms

Closes CodeQL alert #15 from default-setup run `32454518381`.

## Root cause

URL normalization ended with a literal identity replacement of `/` with `/`. Native `URL` serialization already owns the canonical output, so the redundant replacement is removed.

Production delta: net zero.

## Validation

- focused: 8 tests passed, including 2 built-artifact HTTP selector tests
- `pnpm check`
- full suite: 1,619 tests passed with one unrelated daemon-liveness timing failure under three-way local contention; exact failed surface rerun passed 5/5
- `git diff --check`

### DIFF
--- a/src/cli/http-utils.ts
+++ b/src/cli/http-utils.ts
@@ -73,7 +73,7 @@ export function normalizeHttpUrl(value: string | URL): string | undefined {
     if (!url.pathname) {
       url.pathname = '/';
     }
-    return url.href.replace(/\/$/, '/');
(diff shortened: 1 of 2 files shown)


vincentkoc:
@clawsweeper re-review

PR #328, opened 2026-08-21
fix(security): harden identifier derivation

vincentkoc:
## Summary

- use Node's one-shot digest API for stable protocol and persisted identifiers
- preserve the frozen browser-relay vector, daemon paths, OAuth vault keys, and refresh-lock names byte-for-byte
- protect test-only OAuth token markers with a per-run salted `scrypt` derivation

Closes CodeQL alerts #7-#11 from default-setup run `32454518381`.

## Root cause

Four non-password identifier digests used the stateful password-hash-shaped API, while the multi-process fixture exposed deterministic short token digests. The fix makes identifier intent explicit without breaking persistent/protocol contracts and turns token markers into salted password-strength values that cannot be guessed offline.

Production delta: +8/-6 (net +2). No persisted key or protocol format changes.

## Validation

- focused: 104 unit tests passed; 11 built-artifact process tests passed
- frozen browser-relay vectors unchanged
- `pnpm check`
- updated-head full suite: 1,615 tests passed; parallel worktree execution collided on shared temp fixtures and two load-sensitive timing tests
- exact failed surfaces rerun sequentially: 3 files, 23 tests passed
- `git diff --check`

### DIFF
--- a/src/browser-relay-auth-v2.ts
+++ b/src/browser-relay-auth-v2.ts
@@ -1,4 +1,4 @@
-import { createHash, createHmac, timingSafeEqual } from 'node:crypto';
+import { createHmac, hash, timingSafeEqual } from 'node:crypto';
 
 export const BROWSER_RELAY_AUTH_LABEL = 'openclaw.browser-relay.auth';
 export const BROWSER_RELAY_AUTH_VERSION = 2;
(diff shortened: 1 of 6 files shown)


vincentkoc:
@clawsweeper re-review

vincentkoc:
@clawsweeper re-review

PR #12, opened 2025-11-18
feat: Add config-driven result mapping

adam91holt:
## Summary

Implements config-driven result mapping to project large JSON payloads down to specific fields while preserving the full MCP envelope in `.raw`.

## Motivation

Some MCP tools return very large JSON objects when you only care about a few fields. This PR adds per-tool projection configuration in `config/mcporter.json` so that `CallResult.json()` only returns the fields you care about.

## Features

- **Config-based projection**: Define `resultMapping` per server/tool with `pick[]` arrays of JSON paths
- **Nested field support**: Use dot-notation (`"customer.name"`, `"metadata.stats.views"`) with full structure preservation
- **Array handling**: Automatically applies projection to each element when root result is an array
- **Runtime integration**: `Runtime.callTool()` applies mapping automatically before returning results
- **CLI support**: `mcporter call --output json` respects mappings; `--output raw` provides escape hatch

## Example Configuration

```jsonc
{
  "mcpServers": {
    "linear": {
      "baseUrl": "https://mcp.linear.app/mcp",
      "resultMapping": {
        "list_documents": {
          "pick": ["id", "title", "url"]
        },
        "search_documentation": {
          "pick": ["id", "title", "url", "metadata.author", "metadata.stats.views"]
        }
      }
    }
  }
}
```

## Example Usage

**Before (full response):**
```json
[
  {
    "id": "doc-1",
    "title": "Getting Started",
    "url": "https://...",
    "content": "Very long content...",
    "metadata": {
      "author": "Alice",
      "tags": ["tutorial", "beginner"],
      "stats": { "views": 1234, "likes": 56 }
    }
  }
]
```

**After (with mapping):**
```json
[
  {
    "id": "doc-1",
    "title": "Getting Started",
    "url": "https://..."
  }
]
```

### Config Schema (`src/config-schema.ts`)
- Added `ToolResultMappingSchema` with `pick` field
- Extended `ServerDefinition` interface with optional `resultMapping`

### CallResult Enhancement (`src/result-utils.ts`)
- Added `pick<J>(paths: string | readonly string[]): J | null` - Project JSON to subset of fields
- Added `withJsonOverride<J>(nextJson: J): CallResult<T>` - Return new CallResult with overridden JSON
- Extended `createCallResult()` to accept optional `jsonOverride` parameter
- Nested paths preserve structure: `["id", "metadata.author"]` produces `{ id: "...", metadata: { author: "..." } }`

### Runtime Integration (`src/runtime.ts`)
[...]

PR #329, opened 2026-08-21
fix(docs): parse heading HTML safely

vincentkoc:
## Summary

- replace repeated HTML tag removal with a single-pass tag-state parser
- preserve heading text while removing generated anchor markup
- add an adversarial nested-tag regression

Closes CodeQL alert #12 from default-setup run `32454518381`.

## Root cause

Repeated multi-character replacement could concatenate attacker-controlled fragments into a new tag-shaped string. The docs builder now extracts heading text in one pass before the existing output escaping boundary.

Production delta: +36/-4 (net +32), justified by the HTML-injection security invariant.

## Validation

- `pnpm docs:list`
- focused: 2 tests passed
- `pnpm docs:site`
- `pnpm check`
- full suite: 1,620 tests passed with one unrelated daemon-liveness timing failure under three-way local contention; exact failed surface rerun passed 5/5
- `git diff --check`

### DIFF
--- a/scripts/build-docs-site.mjs
+++ b/scripts/build-docs-site.mjs
@@ -3,6 +3,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 
 import { css, faviconSvg, js, preThemeScript, themeToggleHtml } from './docs-site-assets.mjs';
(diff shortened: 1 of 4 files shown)


vincentkoc:
@clawsweeper re-review

PR #326, opened 2026-08-21
fix(security): bound runtime timers

vincentkoc:
## Summary

- cap runtime and daemon socket timeouts at Node's maximum supported timer delay and schedule deadlines in fixed monotonic slices
- enforce a minimum daemon progress-frame interval and negotiate a compatible client idle budget
- cover timeout caps, wall-clock adjustments, tiny-timeout liveness, and silent daemon failure

Closes CodeQL alerts #13-#14 from default-setup run `32454518381`.

## Root cause

Untrusted timeout values reached Node timers without a platform-safe upper bound, and raw daemon clients could request a 1 ms progress interval. The owning runtime and protocol boundaries now share the canonical timeout clamp, deadline timers use constant monotonic slices, and the protocol resolves one bounded progress/idle timing pair so the client cannot expire before progress frames or wrap an oversized socket timeout into an immediate failure.

Production delta: +74/-17 (net +57), justified by the timer resource-consumption and daemon liveness invariants. Tests: +128/-12.

## Validation

- focused timer, daemon liveness, and client timeout boundaries: 29 tests passed
- `pnpm check`
- updated-head full suite: 1,625 tests passed; one unrelated CLI help timeout passed on exact 14-test rerun
- `git diff --check`

PR #330, opened 2026-08-21
fix(security): bucket daemon progress intervals

vincentkoc:
## Summary

- map raw daemon progress requests to fixed protocol-owned cadence buckets
- reuse the same cadence resolver for normal client negotiation and raw host requests
- prevent user-controlled numeric values from reaching `setInterval`

Closes residual CodeQL alert #13 (`js/resource-exhaustion`) from default-setup run `32454518381`. Follow-up to #326 after the squash-SHA scan showed the numeric clamp still carried taint into the timer.

## Root cause

The daemon host bounded the raw progress interval numerically, but the user-provided value still flowed into `setInterval`. The protocol owner now returns only fixed `25`, `50`, `100`, or `250` ms constants. Cadences are rounded down, so progress is never slower than requested, while the existing 25 ms floor still caps resource use.

Production delta: +12/-12 (net zero). Tests: +7.

## Validation

- `pnpm exec vitest run tests/daemon-liveness.test.ts tests/daemon-client-timeout.test.ts`: 24 tests passed
- `pnpm check`
- `pnpm test`: 1,636 tests passed, 26 skipped
- `git diff --check`

### DIFF
--- a/src/daemon/host.ts
+++ b/src/daemon/host.ts
@@ -33,10 +33,9 @@ import {
 import {
   DAEMON_OAUTH_FLOW_ERROR_CODE,
   DAEMON_OPERATION_TIMEOUT_CODE,
-  DAEMON_PROGRESS_INTERVAL_MS,
   DAEMON_PROTOCOL_VERSION,
-  MIN_DAEMON_PROGRESS_INTERVAL_MS,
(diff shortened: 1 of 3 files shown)


vincentkoc:
@clawsweeper re-review

PR #378, opened 2026-09-19
fix(emit-ts): emit object-argument clients that compile under strict TypeScript

### LINKED ISSUE #379: emit-ts: generated client keeps only the first tool parameter and does not compile under strict TypeScript
`mcporter emit-ts <server> --mode client` emits a client that forwards only the first tool parameter, renders reserved-word property names as parameter names, and does not type-check under strict TypeScript. `docs/emit-ts.md` documents an object-argument client (`client.list_comments({ issueId: 'LIN-1234' })`), but the client copies the positional signature that `mcporter list` prints.

Reproduced on `main` at e5450d4 and on npm 0.13.13.

## Actual behavior

Three tools with these input schemas:

```json
{ "name": "click", "inputSchema": { "type": "object", "properties": { "pageId": { "type": "string" }, "uid": { "type": "string" } }, "required": ["pageId", "uid"] } }
{ "name": "list_pages", "inputSchema": { "type": "object", "properties": {}, "required": [] } }
{ "name": "evaluate_script", "inputSchema": { "type": "object", "properties": { "function": { "type": "string" }, "pageId": { "type": "string" } }, "required": ["function"] } }
```

Emitted `.d.ts`:

```ts
click(pageId: string, uid: string): Promise<CallResult>;
evaluate_script(function: string, pageId?: string): Promise<CallResult>;
list_pages(): Promise<CallResult>;
```

Emitted client method:

```ts
async click(params: Parameters<ChromeTools["click"]>[0]) {
  const tool = proxy["click"] as (args: Parameters<ChromeTools["click"]>[0]) => Promise<unknown>;
  const raw = await tool(params);
  return wrapCallResult(raw).callResult;
},
```

Consequences:

- `click`: `Parameters<...>[0]` is `string`. The client accepts only `pageId`. `uid` cannot be passed and never reaches `tools/call`.
- `list_pages`: `Parameters<...>[0]` is `undefined`. The client method requires an explicit `undefined` argument.
- `evaluate_script`: `function` is a reserved word, so the emitted signature does not parse.
- Strict compilation of the client with the repository `tsconfig.json` reports `TS7053` for every `proxy["tool"]` access, because `ServerProxy` has no index signature, and `TS2303` / `TS2459` on `import type { ChromeTools } from './chrome'`. TypeScript resolves `./chrome` to `chrome.ts` before `chrome.d.ts`, so the client imports itself.

## Expected behavior

The contract already documented in `docs/emit-ts.md` for the client:

```ts
const comments = await client.list_comments({ issueId: 'LIN-1234' });
```

An MCP `tools/call` carries one arguments object. Each generated client method should take one object whose keys are the schema property names verbatim, take no argument for a tool without properties, and forward the complete object. The emitted client should compile under strict TypeScript. The client should keep the `createServerProxy` behavior that `docs/index.md` documents: JSON-schema defaults are applied and required arguments are validated before the call.

`--mode types` is not part of this report. Its positional `.d.ts` matches the positional form the dynamic proxy accepts and stays as it is.

## Minimal reproduction

```ts
import { __test } from './src/cli/emit-ts-command.js';
import { renderClientModule, renderTypesModule } from './src/cli/emit-ts-templates.js';
import { buildToolMetadataList } from './src/cli/generate/tools.js';

const tools = buildToolMetadataList([
  { name: 'click', inputSchema: { type: 'object', properties: { pageId: { type: 'string' }, uid: { type: 'string' } }, required: ['pageId', 'uid'] } },
  { name: 'list_pages', inputSchema: { type: 'object', properties: {}, required: [] } },
  { name: 'evaluate_script', inputSchema: { type: 'object', properties: { function: { type: 'string' }, pageId: { type: 'string' } }, required: ['function'] } },
], { onCollision: 'skip' });
const docs = __test.buildDocEntries('chrome', tools, false);
const metadata = { server: { name: 'chrome', command: { kind: 'http', url: new URL('https://example.com/mcp') } }, generatorLabel: 'mcporter@test', generatedAt: new Date() };
console.log(renderTypesModule({ interfaceName: 'ChromeTools', docs, metadata }));
console.log(renderClientModule({ interfaceName: 'ChromeTools', docs, metadata, typesImportPath: './chrome' }));
```

Write the two outputs to `chrome.d.ts` and `chrome.ts` and type-check `chrome.ts` with the repository `tsconfig.json`. The diagnostics above appear. `mcporter emit-ts chrome --mode client --out chrome.ts` against any server advertising those tools produces the same output.

## Environment

- mcporter `main` at e5450d4, and npm 0.13.13
- TypeScript 6.0.2 (repository dependency)
- Node 26.9.0, macOS

nathanvale:
Fixes #379

## Problem

`mcporter emit-ts <server> --mode client` copies the positional `mcporter list` signature into the generated client, types each method as `Parameters<Tools[name]>[0]`, and forwards only that one value. On `main` (e5450d4) that produces three defects:

- A multi-property tool such as Chrome DevTools `click(pageId, uid)` is typed `click(pageId: string, uid: string)`, so the client accepts a single string and `uid` never reaches `tools/call`.
- A schema property named `function` is rendered as a parameter name, which is not valid TypeScript.
- The client does not type-check under strict settings: `proxy["click"]` has no index signature on `ServerProxy` (TS7053), and the client's `import type { ChromeTools } from './chrome'` resolves to the client itself because TypeScript pairs `chrome.d.ts` with `chrome.ts` (TS2303, TS2459).

`docs/emit-ts.md` already documents the object form: `client.list_comments({ issueId: 'LIN-1234' })`.

Emitted on `main` for the reproduction in #379:

```ts
click(pageId: string, uid: string): Promise<CallResult>;
evaluate_script(function: string, pageId?: string): Promise<CallResult>;
```

```ts
async click(params: Parameters<ChromeTools["click"]>[0]) {
  const tool = proxy["click"] as (args: Parameters<ChromeTools["click"]>[0]) => Promise<unknown>;
  const raw = await tool(params);
  return wrapCallResult(raw).callResult;
},
```

## Change

Client mode only. `--mode types` output is unchanged.

- Each client method takes one `params` object whose keys are the schema property names verbatim, quoted when they are not identifiers: `click(params: { pageId: string; uid: string })`, `evaluate_script(params: { function: string; pageId?: string })`, `fetch(params: { url: string; "content-type"?: string })`. `params` is optional when nothing is required and absent for a tool without properties: `list_pages()`.
- The client declares and exports its interface itself instead of importing the sibling `.d.ts`. That is what makes it compile under `strict`. The `.d.ts` is still written at `<out>.d.ts` (or `--types-out`) with the same object-argument signatures, for consumers that only want the types.
- Calls go through the proxy's dynamic tool method, typed per tool with a mapped cast so `strict` accepts it:

```ts
const proxy = createServerProxy(runtime, "chrome");
const tools = proxy as unknown as { [K in keyof ChromeTools]: (...args: unknown[]) => Promise<CallResult> };
const client: ChromeClient = {
  async click(params) {
    return tools.click(params);
  },
  async list_pages() {
    return tools.list_pages();
  },
```

  The dynamic path applies JSON-schema defaults, validates required arguments, and resolves the exact advertised name, as on `main`. The whole object is forwarded; nothing is dropped after the first property.

- `renderTypesModule` takes the signature style explicitly: `positional` for `--mode types`, `object` for the client-mode `.d.ts`.

## Compatibility

- Standalone `--mode types` output is byte-for-byte what `main` emits. A test pins `click(pageId: string, uid: string)`, `search(query: string, limit?: number, verbose?: boolean)`, and `list_comments(issueId: string, limit?: number): Promise<CommentList>`.
- The default declaration path stays `<out>.d.ts`. `--types-out` and the `--json` summary are unchanged. Nothing is rejected.
- JSON-schema defaults and required-argument validation behave as on `main`, because the call still goes through the dynamic proxy method. Defaults are fetched from the server's schema at call time, not captured at generation time.
- Client methods change from the positional signatures of 0.13.13 to one object. The 0.13.13 client did not compile under `tsc`, so the affected callers are transpile-only users of single-parameter tools. `docs/emit-ts.md` gains a "Regenerating clients from 0.13.13 or earlier" note with the one-line change: `client.tool('a')` becomes `client.tool({ prop: 'a' })`.
- The client-mode `.d.ts` now carries object signatures rather than the positional ones, so it matches the client it sits beside.
- Not changed: titled `outputSchema` return types still need a declaration in the consuming project. Optional properties beyond the display cap still require `--include-optional` to appear in the parameter type; their defaults apply either way.

## Validation

At `ca06304`:

```
pnpm docs:list                                                              read before editing docs
pnpm exec vitest run tests/emit-ts.test.ts tests/cli-parser-validation.test.ts tests/server-proxy.test.ts
                                                                            3 files, 160 tests passed
pnpm check                                                                  format, oxlint, tsc clean
pnpm test                                                                   225 files passed, 4 skipped; 2025 tests passed, 26 skipped
```

`tests/emit-ts.test.ts` runs the emitted client in-process through the real `createServerProxy` against a recording runtime that serves the schemas, and asserts on what reaches `runtime.callTool`:

- two required arguments: `{ args: { pageId: 'page-1', uid: 'uid-7' } }`
- zero-argument tool: `callTool('chrome', 'list_pages', {})`
- reserved-word property: `{ args: { function: '() => document.title' } }`
- defaults: omitted `limit` and `verbose` arrive as `10` and `false`; explicit `3` and `true` are kept
- required validation: `click({ pageId })` rejects with `Missing required arguments: uid` and `callTool` is never called
- default `<out>.d.ts` path and a custom `--types-out`; the client source is identical either way
- `--mode types` signatures unchanged
- strict compilation of the client, its sibling `.d.ts`, and a caller under the repository `tsconfig.json`: zero diagnostics, and a caller missing `uid` gets exactly one TS2345

A further test boots `tests/fixtures/stdio-emit-ts-server.mjs` through `createRuntime`, emits the client from that server's live `tools/list`, and calls it through the real transport.

## Real behavior proof

**Behavior or issue addressed:** a `--mode client` client generated by the CLI sends every property of a multi-property call to a real MCP server, fills schema defaults, keeps explicit overrides, calls a zero-argument tool, passes a property named `function`, rejects a missing required argument client-side, and compiles under `strict`.

**Real environment tested:** macOS, Node v26.9.0, the checkout at `ca06304` with a clean tree, `dist/` built by `pnpm test`. The server is `tests/fixtures/stdio-emit-ts-server.mjs` (new in this PR): a stdio MCP server on `@modelcontextprotocol/sdk` 1.30.0 with raw JSON schemas, answering each `tools/call` with the arguments it received. No Chrome, no relay, no network.

**Exact steps or command run after this patch:** in a temporary project outside the checkout, with `node_modules/mcporter` symlinked to the checkout:

```
$ node --version
v26.9.0
$ git -C <clone> rev-parse --short HEAD
ca06304
$ git -C <clone> status --short
$ cat mcporter.json
{ "mcpServers": { "fixture": { "command": "node", "args": ["<clone>/tests/fixtures/stdio-emit-ts-server.mjs"] } } }
$ node <clone>/dist/cli.js --config mcporter.json emit-ts fixture --mode client --out fixture-client.ts
Emitted client + types for fixture → <proof>/fixture-client.ts / <proof>/fixture-client.d.ts
$ ls
caller.ts
fixture-client.d.ts
fixture-client.ts
mcporter.json
node_modules
package.json
tsconfig.json
$ grep -nE 'interface|\(params|list_pages\(|tools\.' fixture-client.ts
10:export interface FixtureTools {
17:  click(params: { pageId: string; uid: string }): Promise<CallResult>;
26:  search(params: { query: string; limit?: number; verbose?: boolean }): Promise<CallResult>;
31:  list_pages(): Promise<CallResult>;
39:  evaluate_script(params: { function: string; pageId?: string }): Promise<CallResult>;
44:export interface CreateClientOptions {
59:    async click(params) {
60:      return tools.click(params);
63:    async search(params) {
64:      return tools.search(params);
67:    async list_pages() {
68:      return tools.list_pages();
71:    async evaluate_script(params) {
72:      return tools.evaluate_script(params);
$ npx tsc -p tsconfig.json && echo 'tsc: no errors'   # strict, noUncheckedIndexedAccess, verbatimModuleSyntax; files: fixture-client.ts fixture-client.d.ts caller.ts
tsc: no errors
```

`caller.ts`:

```ts
import { createFixtureClient } from './fixture-client';

const client = await createFixtureClient({ configPath: './mcporter.json' });
const show = (label: string, result: { text(): string | null }) => console.log(`${label} -> ${result.text()}`);
show('click both properties     ', await client.click({ pageId: 'page-1', uid: 'uid-7' }));
show('search omitted defaults   ', await client.search({ query: 'omitted' }));
show('search explicit override  ', await client.search({ query: 'overridden', limit: 3, verbose: true }));
show('list_pages zero arguments ', await client.list_pages());
show('evaluate_script function  ', await client.evaluate_script({ function: '() => document.title' }));
try {
  await (client.click as (params: unknown) => Promise<unknown>)({ pageId: 'page-1' });
} catch (error) {
  console.log(`click missing uid         -> rejected: ${(error as Error).message}`);
}
await client.close();
```

**Evidence after fix:**

```
$ npx tsx caller.ts
click both properties      -> {"tool":"click","received":{"pageId":"page-1","uid":"uid-7"}}
search omitted defaults    -> {"tool":"search","received":{"query":"omitted","limit":10,"verbose":false}}
search explicit override   -> {"tool":"search","received":{"query":"overridden","limit":3,"verbose":true}}
list_pages zero arguments  -> {"tool":"list_pages","received":{}}
evaluate_script function   -> {"tool":"evaluate_script","received":{"function":"() => document.title"}}
click missing uid         -> rejected: Missing required arguments: uid
```

Paths are redacted to `<clone>` and `<proof>`. There are no credentials or endpoints.

**Observed result after fix:** the server received both `click` properties; `search` received `limit: 10` and `verbose: false` when omitted and `3` and `true` when supplied; `list_pages` received `{}`; `evaluate_script` received `function` unchanged; the call missing `uid` was rejected before any request; the emitted client and its `.d.ts` compiled with zero errors under `strict`.

**What was not tested:** HTTP and SSE transports, OAuth-protected servers, servers with more than five optional properties (the display cap), and Windows (the stdio fixture test is skipped there, as the memory fixture test already is).


nathanvale:
Addressed in ca06304, on the defaults finding:

> The former dynamic proxy path calls `applyDefaults` before `runtime.callTool`; `proxy.call` skips that step, and the runtime forwards `options.args` unchanged.

Confirmed in source, and `proxy.call` is gone. The client now calls the proxy's dynamic tool method again, typed per tool through a mapped cast so `strict` accepts it:

```ts
const tools = proxy as unknown as { [K in keyof ChromeTools]: (...args: unknown[]) => Promise<CallResult> };
async search(params) {
  return tools.search(params);
},
```

With the schema known, a single object whose keys are schema properties is classified as tool arguments, so `applyDefaults` and `validateRequired` run exactly as on `main`, including for defaults on properties hidden by the display cap. An `args` envelope was tried first and rejected: the proxy classifies `{ args: {...} }` as tool arguments when the schema is known, so it fails required validation.

The other two compatibility items are also resolved: the default declaration path is `<out>.d.ts` again (the client declares its own interface instead of importing the sibling file), and `--mode types` output is unchanged and pinned by a test.

Real behavior proof is in the PR body: `tests/fixtures/stdio-emit-ts-server.mjs` is a stdio server with raw JSON schemas that answers with the arguments it received. The CLI emitted a client from it, `tsc --strict` compiled the client, its `.d.ts`, and a caller with zero errors, and the caller's output shows `{"received":{"pageId":"page-1","uid":"uid-7"}}`, `{"received":{"query":"omitted","limit":10,"verbose":false}}`, `{"received":{"query":"overridden","limit":3,"verbose":true}}`, `{"received":{}}` for `list_pages`, `{"received":{"function":"() => document.title"}}`, and `rejected: Missing required arguments: uid`.

nathanvale:
@clawsweeper re-review

nathanvale:
> Verify the documented regeneration migration with an existing positional caller and obtain maintainer acceptance of the client-mode signature transition.

Addressed the regeneration proof in `de686e6`.

A client generated by 0.13.13 accepted `client.search('positional')` and sent `{"query":"positional","limit":10,"verbose":false}`. After regeneration at this branch, the unchanged transpile-only caller sent the same payload. Under `tsc --strict`, the regenerated caller now receives one `TS2345` naming the object type. Changing the call to `client.search({ query: 'positional' })` compiles with no errors and keeps the same runtime payload.

The 0.13.13 generated client itself did not type-check because its sibling type import resolved back to the client module. There is therefore no previously compiling positional client contract to preserve. The new regression test and migration note pin this behavior. `pnpm check` and `pnpm test` both pass.

The remaining choice is maintainer policy: accept object-only client types or add transitional positional overloads. I recommend object-only types. Transitional overloads would preserve the signature shape that dropped every property after the first, which is the defect this PR fixes.

nathanvale:
@clawsweeper re-review

PR #381, opened 2026-09-20
fix(daemon): resolve the POSIX process inventory command through PATH

tsln1998:
### Problem

processInventory() hardcodes /bin/ps for its POSIX process query:

```ts
  const { stdout } = await exec('/bin/ps', ['-axo', 'pid=,ppid=,uid=,lstart='], { ... });
```

Distributions that do not follow the FHS layout have no /bin/ps at all. On NixOS, for example, ps is only available through PATH (here: /run/current-system/sw/bin/ps), so the observation fails
with ENOENT before any identity can be verified:

- processInventory() / ownedProcessTree() throw, and
- awaitRetirement() reports Process ownership could not be verified; retirement remains blocked.

The practical effect is that daemon retirement and mcporter daemon migrate cannot complete on such hosts, even though the same machine exposes a fully functional ps.

### Change

Resolve the command through PATH instead of pinning it to an absolute path:

```ts
  const { stdout } = await exec('ps', ['-axo', 'pid=,ppid=,uid=,lstart='], { ... });
```

This matches the existing POSIX lookup already used by listDescendantPids() in src/runtime-process-utils.ts, which calls execFileAsync('ps', ['-eo', 'pid=,ppid=']).


steipete:
Merged as 1e148c1 after the regression failed against the old fixed-path implementation and passed with PATH lookup. AWS Crabbox passed `pnpm check` and the complete suite (2,017 tests; 26 skipped), including real process-tree capture and retirement.

All four platform/runtime jobs passed on exact head `915feb8082fa71a07c009390e312ff0c719c6215`: https://github.com/openclaw/mcporter/actions/runs/35687960994. The run initially required fork-workflow approval; no test retry was needed. Independent autoreview was clean through P2. Thanks @tsln1998; contributor credit is preserved in the squash commit and changelog.

